PR #9, opened 2024-10-29
Stat lib overwite

FlyingFish800:
Added an option to overwrite a stat if it exists already exists. The new stat's own value is also allowed to be used in the new calculation (Ex: test=test+1). This functionality is disabled by default to prevent accidental changes of a stat, and can be enabled with the 'overwrite' parameter of derive_stat.


5surim:
Hi @FlyingFish800 ,

could you clean up the commits?
This PR should have only one commit [5aec048].

Thanks,
Surim

FlyingFish800:
Added write protect for non-derived stats, and squashed commits in [new pr](https://github.com/Litz-Lab/Scarab-infra/pull/12).

PR #10, opened 2024-11-04
Geekbench Dockerfile Env Setting

yzhao321:
Before build the images, the license key should be copied into the current dir.


5surim:
Hi @yzhao321 

thanks for the update on the geekbench. Could you also update the BINCMD for geekbench and provide the perf results in this PR?

Surim

PR #13, opened 2024-11-14
Fixed issues due to existance of metadata write_protect column, and t…

FlyingFish800:
Fixed issue where the write_protect column's boolean values were seen as unique experiment/workload/config names. Also disabled echo for serve_jupyter.sh script as it makes it harder to use.


5surim (approved):
LGTM

PR #16, opened 2024-12-11
Clean up scarab stats library and refactor APIs for common use cases and better color/hatch sets

5surim:
Clean up scarab stats library and refactor APIs for common use cases and better color/hatch sets.
Examples of new API usage are available in scarab_stats_quick_start.ipynb

plot_workloads() plots only one stat but all the workloads and configurations (multiple stats within a single plot are usually not easy to read because of a conflict on y-axis units)
plot_workloads_speedup() plots speedups of only one stat but all the workloads and configurations
plot_stacked() plots the stacked bars of raw counters of given stats with better legends of configurations x stats
plot_stacked_fraction() plots the stacked bars of the ratio of given stats


FlyingFish800 (dismissed):
For plot_workloads and plot_workloads_speedup I think it would be better to accept a string for the stat instead of a list if we can only use one. Also for plot_workloads_speedup, I don't think the speedup_baseline parameter should have a default value.

I tested all the changes and verified everything works, and the new graphs look good!

5surim:
@FlyingFish800 I agree. Will update this PR! Thanks a lot for reviewing this!

5surim:
@FlyingFish800 To reuse `experiment.retrieve_stats`, let's leave it as it is for now. Just having the warning should be fine. Could you approve this change again?

FlyingFish800 (dismissed):
LGTM

yql5510718 (approved):
LGTM

PR #17, opened 2024-12-13
remove disabling ASLR from docker build as it faults

kofyou:
Disabling ASLR while building an image is not possible; containers share the setting with the host: https://stackoverflow.com/questions/54845095/cannot-run-sysctl-command-in-dockerfile

This commit turns off ASLR before running the simpoint script; the user needs to recover the setting afterwards manually.


5surim (approved):
LGTM

PR #20, opened 2025-01-09
Create docker-image.yml for a GitHub workflow to build the docker images

5surim:
A workflow for a prebuilt docker image. This workflow creates a docker image named 'allbench_traces:<githash>'.


5surim:
@hlitz Could you confirm that GitHub Packages is enabled for this repository (Scarab-infra) or organization (Litz-Lab)?:

Go to your repository or organization settings.
Navigate to Packages and ensure ghcr.io is enabled.

.github/workflows/docker-image.yml
```diff
@@ -0,0 +1,48 @@
...
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Set timestamp
+        id: timestamp
+        run: echo "timestamp=$(date +%s)" >> $GITHUB_ENV
+
+      - name: Build Docker images
+        run: |
+          docker build . --file ./allbench_traces/Dockerfile --no-cache --tag ghcr.io/litz-lab/scarab-infra/allbench_traces:${{ env.timestamp }}
+          docker build . --file ./docker_traces/Dockerfile --no-cache --tag ghcr.io/litz-lab/scarab-infra/docker_traces:${{ env.timestamp }}
```
hlitz, line 38: are we merging the two into a single image referencing a path to the trace files?

common/Dockerfile.common
```diff
@@ -81,8 +81,5 @@ RUN cd $tmpdir/ && \
     make -C SimPoint.3.2 && \
     ln -s SimPoint.3.2/bin/simpoint ./simpoint
 
-# disable ASLR
```
hlitz, line 84: why are we removing this? Should it go into a different commi?

hlitz (changes requested):
generate only one docker image

.github/workflows/docker-image.yml
5surim: Yes, I will fix this as we discussed today.

common/Dockerfile.common
5surim: It has already been removed from another PR. Just rebased 'main' so that we have this change for a successful docker build.

hlitz (approved):
LGTM

PR #22, opened 2025-01-21
Refactor the whole infrastructure and documentation

5surim:
This PR includes refactoring of the whole infrastructure and a simplified main README.md. The main purpose of this refactoring is to make 'build' and 'run' docker independent from other runs (simpoint/trace/scarab). We want to split 'builld' and 'run' and make 'allbench_traces' and 'docker_traces' merged and the trace option configurable during 'run'.
We don't have a good reason to maintain the functionality to build multiple images or to run multiple containers at once. It can be manually done or automatically done by Slurm.

The following are the main changes.
1. Split 'docker build' and 'docker run' within the script (new -b -r options)
2. Remove 'apps.list' as an input list of workloads. This was needed when we wanted to do everything (build a docker image, run a container, run simpoint/tracing, run scarab) at once to provide a list of workloads. 
3. The list of workloads needed for tracing/simpointing should be provided with -w.
4. Building image only if the current git status does not have any uncommitted changes for an up-to-date image and only if the same githash version of image does not exist in the system.
5. Reorganize all the files for a better readability.


5surim:
@hlitz, @FlyingFish800 This PR is ready for review.

README.md
```diff
@@ -1,150 +1,80 @@
...
-cd /home/$USER/docker_traces/scarab
-(update scarab)
+cp json/exp.json json/your_experiment.json
 ```
-##### To build scarab again with the updated scarab, go inside the container and build.
+2. Edit your_experiment.json to describe your experiment. Please refer to json/exp.json for the describtion. To set the 'workload manager' = slurm, Slurm should be installed.
+3. Run all experiments
 ```
-docker exec --user=$USER --privileged docker_traces_$USER /bin/bash -c "cd /home/$USER/scarab/src && make"
+./run.sh -s your_experiment
```
hlitz, line 41: I find the arguments non-intuitive and it will require running ./run -h frequently.
 -s = run -i = status ..
I wonder if we should do --status, --list, --submit, --kill, --build, --clean
instead

README.md
```diff
@@ -1,150 +1,80 @@
...
+./run.sh -k your_experiment
 ```
 
-#### Check the built image
+## Open an interactive shell of a docker container for the purpose of debugging/development
 ```
-docker images
-REPOSITORY                   TAG       IMAGE ID       CREATED        SIZE
-example                      latest    1dd7a6097ef0   3 hours ago    6.66GB
+./run.sh -l your_experiment
```
hlitz, line 57: What is the dependency between interactive mode and your_experiment ? The image file path only?

json/exp.json
```diff
@@ -1,5 +1,9 @@
 {
+    "_architecture": "one of the available architectures (PARAMS.<architecture>) found in scarab repository",
```
hlitz, line 2: rephrase: "The path to a PARAMS.<architecture> file which defines the machine parameters of the default configuration."

json/exp.json
```diff
@@ -1,5 +1,9 @@
...
     "architecture": "sunny_cove",
+    "_workload_group": "one directory name in workloads",
+    "workload_group": "allbench_traces",
```
hlitz, line 5: what is a workload group and why is it needed in addition to workload list? 
what if we want to run multiple workload groups?
why can't we derive workload_group from the workload? (maybe in the workload database we were discussing we can just add that info to each workload)

I am ok submitting the patch as is and then refactor as soon as we have the workload database.

json/exp.json
```diff
@@ -9,8 +13,19 @@
         "verilator",
         "xgboost"
     ],
-    "experiment": "slurm_experiment",
-    "docker_home": "/<path>/<to>/slurm_home",
+    "_experiment": "experiment name",
```
hlitz, line 16: rephrase: "The name of the experiment. scarab-infra will create a new folder named <experiment> containing all files needed to execute the simulation in the FIXME directory. The <experiment> folder will also contain all generated simulation statistics/outputs."

json/exp.json
```diff
@@ -9,8 +13,19 @@
...
+    "experiment": "exp",
+    "_simulation_mode": "1: execution-driven simulation w/o SimPoint, 2: trace-based simulation w/o SimPoint, 3: execution-driven simulation w/ SimPoint, 4: trace-based simulation w/ SimPoint, 5: PT-trace-based simulation w/ SimPoint",
```
hlitz, line 18: ok to submit as is, but for the future we should consider retrieving this data from the workload database.
Also, we might want to run a mix of PT, memtrace, exec driven apps as part of the same experiment.

json/exp.json
```diff
@@ -9,8 +13,19 @@
...
+    "simulation_mode": "4",
+    "_workload_manager": "manual: run each exp scenario with all the simpoints at a time, slurm: Slurm schedules simulations based on its policy",
+    "workload_manager": "manual",
+    "_root_dir": "/<path>/<to>/<docker_home>",
```
hlitz, line 22: is this the root dir that is mounted in the docker container? If yes, rephrase:
"Root directory mounted into the docker container, for instance, the user's home directory."

json/exp.json
```diff
@@ -9,8 +13,19 @@
...
+    "root_dir": "/home/$USER/allbench_home",
+    "_scarab_path": "/<path>/<to>/<scarab>/<repo>",
```
hlitz, line 24: rephrase: "The path to the scarab repository that contains the scarab binary used to run the experiment."

json/exp.json
```diff
@@ -9,8 +13,19 @@
...
+    "scarab_path": "/home/$USER/src/scarab",
+    "_simpoint_traces_dir": "/<path>/<to>/<simpoints_traces>",
```
hlitz, line 26: rephrase: "The local directory containing all simpointed traces."

hlitz (commented):
Looks great. I have only reviewed README and exp.json

json/exp.json
5surim: Yes, this PR is not constructing the database and merging PT/memtrace.
I will create both issues for the following PRs.

json/exp.json
5surim: For now, this infra is working well for trace-based simulation.
All the trace-based simulations fall through the same workload group, and we have not considered different workload groups so far.
We can convert this to the list of key values (group - list of workloads) to enable exec-driven simulations with different workload groups. I would include this feature in the exec-driven issue, not this PR.
We still need this workload group to identify if it is an exec-driven simulation or trace-based simulation even with the database because each workload group might have the workload with the same name. For example, exec-driven mode has <sysbench>:<mysql> and trace-based mode has <allbench>:<mysql>.

json/exp.json
5surim: We just take the name of the architecture, not the path because we should be able to get the path from <scarab_path>

README.md
5surim: -s = simulation -i = info
-r = run an interactive shell of docker container.
I agree. I will just remove the short options.

README.md
5surim: We want to create an interactive shell of a container where we are able to run a debug mode of simulation of a workload. We need to specify which workload/group we want to simulate where it choose the corresponding docker image to start a container.

5surim:
Can we merge this to the main for further PRs?

hlitz (approved):
LGTM, small remaining issues can be fixed in future PRs

json/exp.json
5surim: The configurations described in the json will always overwrite the parameters on top of the base PARAMS.<architecture>.
If we decouple the PARAMS file from the scarab, I think it will even cause inconsistencies and confusion. For example, we expect that the simulation runs with the given PARAMS file, but if the user doesn't empty the configurations, then the parameters are overwritten, and the simulation is not exactly what the user expects.
I think the only case where we want to directly use the PARAMS file is when we debug a single scenario. In that case, the simulation results should already exist, and the user should open an interactive shell of the container and run the simulation by using the PARAMS file.

PR #23, opened 2025-01-23
Refactoring readme

hlitz:


hlitz:
Not really a PR. Just my thoughts of how a simplified scarab-infro interface could look like.

PR #42, opened 2025-02-11
Add printing trace stack on an exception

5surim:
The full stack print on an exception will ease debugging.


FlyingFish800 (approved):
LGTM!

PR #45, opened 2025-02-12
Use a prebuilt DockerHub image instead of building locally

moonseoppkim:
Skipping local docker image build process, if we found a pre-build docker image which is matching to the commit hash from github package(automatically generated image based on the commit hash).

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -94,6 +94,18 @@ build () {
     exit 1
   fi
 
+  # Local image not found. Trying to pull pre-built image from GitHub Packages...
+  if docker pull ghcr.io/litz-lab/scarab-infra/"$APP_GROUPNAME:$GIT_HASH"; then
+    echo "Successfully pulled pre-built image."
+    docker tag ghcr.io/litz-lab/scarab-infra/"$APP_GROUPNAME:$GIT_HASH" "$APP_GROUPNAME:$GIT_HASH"
+    echo "Tagged pulled image as $APP_GROUPNAME:$GIT_HASH"
+    docker rmi ghcr.io/litz-lab/scarab-infra/"$APP_GROUPNAME:$GIT_HASH"
+    exit 1
+  else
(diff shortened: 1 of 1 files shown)


run.sh
```diff
@@ -94,6 +94,18 @@ build () {
     exit 1
   fi
 
+  # Local image not found. Trying to pull pre-built image from GitHub Packages...
+  if docker pull ghcr.io/litz-lab/scarab-infra/"$APP_GROUPNAME:$GIT_HASH"; then
+    echo "Successfully pulled pre-built image."
+    docker tag ghcr.io/litz-lab/scarab-infra/"$APP_GROUPNAME:$GIT_HASH" "$APP_GROUPNAME:$GIT_HASH"
+    echo "Tagged pulled image as $APP_GROUPNAME:$GIT_HASH"
+    docker rmi ghcr.io/litz-lab/scarab-infra/"$APP_GROUPNAME:$GIT_HASH"
```
5surim, line 102: Just a quick question. Is the tagged image still alive when you remove the original image?

5surim (commented):
Hi @moonseoppkim ,

Thanks for the PR!
I just have a quick question. Could you answer that?

The rest looks good to me.

Surim

run.sh
moonseoppkim: Yes, tagged image still alive.

REPOSITORY                                      TAG       IMAGE ID       CREATED          SIZE
allbench_traces                                 8a0131a   98bbddbdeb8e   42 minutes ago   3GB

commit 8a0131a03646f397c3f23e832a0291a327f1e1cb (HEAD -> prebuilt_dockerimg, origin/prebuilt_dockerimg)
    Use a prebuilt DockerHub image instead of building locally

This is the one I made by using this script at the LADP server.

5surim: Great!

5surim (approved):
LGTM

PR #46, opened 2025-02-13
Fix for missing utility API

5surim:
set_app_bincmd is used by both run.sh and run_single_simpoint.sh currently


FlyingFish800 (approved):
LGTM!

PR #65, opened 2025-03-05
Add scarab build option to trace/simulation JSON

5surim:
- A scarab build option is added into JSON.
- remove some ugly indentations in the JSON files
- fix a wrong export (PIN_ROOT) in user_entrypoint.sh
- Add gdb installed for an interactive shell of a container for the purpose of debugging


FlyingFish800 (approved):
LGTM!

PR #75, opened 2025-03-18
Small Fixes to Slurm Runner

naomirehman1008:
Previous logic for checking which image is available on a node did not detect that it was not available, causing runs to fail on that node. Logic for checking which simulations are running always printed running even if no simulations were running.


5surim (approved):
LGTM!

PR #80, opened 2025-03-21
Fix features:  typos, errors(divide by zero), NaN handling

Kiminje:
Add features: apply parallelism, optimize codes, add save points + support continuous update file

Handling divided by zero.
Some stats have Nan values, so it can cause dbz errors when you use derive_stats. It simply inserts zero when the dbz flag is set.

Parallelism
Since we don't have to care about the order of stats and simpoints, it can be processed in parallel.

Optimizations (x8 faster when you set 40 workers)
Shared state management:
Uses multiprocessing.
Manager to create shared data structures: 
shared_results: Collects results from all threads
completed_groups: Tracks progress across threads
errs: Tracks error conditions
lock: Ensures thread-safe updates to shared resources. 

Efficient result collection: Results are first collected in the shared list
After all processing is complete, a single DataFrame update is performed, avoiding repeated DataFrame concatenation operations, which are expensive. 
Progress monitoring:
Implements a progress tracking system (completed_groups.value). Provides periodic updates (every 10 groups)
Reports percentage completion. 
Resource cleanup:
Explicit cleanup of manager resources
It uses a final block to ensure cleanup happens even if errors occur. Clears references to managed objects before shutting down the manager

Save data frame
Now, after loading the data frame, it will be saved as a pkl file. So whenever after saving df, there are no repeating load operations unless the code detects new configurations or workloads. If there are new configs or workloads, the code generates df for incoming data and attaches it to the existing df.

### DIFF
--- a/scarab_stats/scarab_stats.py
+++ b/scarab_stats/scarab_stats.py
@@ -230,29 +230,6 @@ class Experiment:
         # Set the row in the dataframe
         self.data.loc[insert_index] = row
         
-        # Check for NaN values and replace with zeros while reporting
-        row_data = self.data.loc[insert_index]
-        nan_count = row_data.isna().sum()
-        
-        if nan_count > 0:
-            # Report division by zero but continue processing
-            print(f"INFO: Detected {nan_count} NaN values in result of '{equation}'")
-            print(f"INFO: These are likely due to division by zero. Replacing NaNs with 0.")
-            
-            # Replace NaN values with zeros
-            self.data.loc[insert_index] = self.data.loc[insert_index].fillna(0)
-            
-            # Print the locations where NaNs were detected for debugging
-            nan_locations = []
-            for col in self.data.columns[3:]:  # Skip the first 3 metadata columns
-                if pd.isna(row_data[col]):
-                    nan_locations.append(str(col))
-            
-            if len(nan_locations) > 5:
-                print(f"INFO: First 5 locations with NaN values: {', '.join(nan_locations[:5])}...")
-            else:
-                print(f"INFO: Locations with NaN values: {', '.join(nan_locations)}")
-        
         return
 
     def to_csv(self, path:str):
@@ -436,7 +413,6 @@ class Experiment:
                 manager.shutdown()
                 
                 # Additional sleep to allow for cleanup
-                import time
                 time.sleep(0.1)
             except Exception as e:
                 print(f"WARNING: Non-critical error during resource cleanup: {str(e)}")
@@ -1279,7 +1255,7 @@ class stat_aggregator:
         all_data = experiment.retrieve_stats(configs, stats, workloads, aggregation_level="Config")
         if speedup_baseline != None: baseline_data = experiment.retrieve_stats([speedup_baseline], stats, workloads, aggregation_level="Config")
 
-        plt.figure(figsize=(6+num_workloads,8))
+        plt.figure(figsize=(6+len(configs),8))
 
         # For each stat
         for x_offset, stat in enumerate(stats):
--- a/scarab_stats/scarab_stats.py
+++ b/scarab_stats/scarab_stats.py
@@ -14,6 +14,11 @@ import sys
 sys.path.append("../") # go to parent dir
 import scripts.utilities as utilities
 importlib.reload(utilities)
+import concurrent.futures
+import multiprocessing
+import time
+import pickle
+import hashlib
 
 def get_elem(l, i):
     return list(map(lambda x:x[i], l))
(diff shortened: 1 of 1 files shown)


5surim:
Hi @Kiminje, thanks for the patch.

What is the difference between the pkl file and the csv result from our legacy save_experiment API?
The csv needs to load the experiment again when we have new configurations, but your pkl does not?
Can you share the execution time to load the experiment when you run the whole spec2017 integer subsuite?
What is the size of pkl file (also for the whole spec2017 integer subsuite)?

@FlyingFish800, Can you test this patch and see if there is any incorrectness in the derived/distribution stats? 
Do you think we no longer need the csv, or do we still need it for a human-readable stat file?

Surim

5surim:
> @5surim 5 min 38sec for processing rate_int(if pkl exists, it takes 3.7 sec to load). 2 mins for collecting metadata for every simpoints and config. Then it calls calculate_distribution_stats, which took 3 mins Also, it doesn't need a csv file to load the data frame. With using pkl, only new configurations are processed and attached to the existing data frame.

Are you saying the existing csv saved by `save_experiment` needs to load everything when you add some new configurations?

5surim:
@Kiminje What is dbz error exactly?

Kiminje:
I think we don't need csv anymore. But it requires additional features removing deprecated configs

5surim:
> @5surim
> 
> 1. Yes, but pkl doesn't require converting a csv file to a data frame, so it's always faster than loading a csv file.
> 2. dbz refers to 'divided by zero'. When you use custom stats using derive_stats, which contains a divider, Nan values in your stats can cause this issue.

1. Is pkl human-readable?
2. Showing up 'divided by zero' is essential because it could imply another potential issue not the API itself. We don't even want to calculate a derived stat if there is a NaN value. Removing an error by ignoring NaN is not a solution. Please figure out which stat you are trying to derive and why you are deriving the stat even though the value is NaN.

Kiminje:
1. no, do we need to leave csv functions for debug?
2. when calculating the hit rate for each component, 0 counts can exist.
I'll add to control whether allow dbz or not

5surim:
> 1. no, do we need to leave csv functions for debug?
> 2. when calculating the hit rate for each component, 0 counts can exist.
>    I'll add to control whether allow dbz or not

2. Please provide your precise equation for your derived stat and the exact name of the stat. We need to distinguish the real zero and NaN. We never want to allow dbz.

Kiminje:
The calculation can cause an issue when generating a hit rate for each TAGE component if the workloads are not predicted by a specific component. This is also can happen when comparing predictors with different components
image
29M of pkl file stores two configs for spec2017_rate_int. 
image
Each configuration has 365M (total 730M)
image

5surim:
> The calculation can cause an issue when generating a hit rate for each TAGE component if the workloads are not predicted by a specific component. This can also happen when comparing predictors with different components. HIT_RATE = CORRECT / (CORRECT + MISPREDICT) but what if CORRECT and MISPREDICT are both zero? That's why we need a boolean argument for derive_stats to decide whether to allow dbz or not. image 29M of pkl file stores two configs for spec2017_rate_int. image Each configuration has 365M (total 730M) image

I'm pretty sure we already have MANY other stats with this rate calculation. Heiner, Alex, and I have already discussed this a lot in many meetings.
Do you see both CORRECT and MISPREDICT for all the sim points?
so hit_rate = (0*seg1 + 0*seg2 + 0*seg3) / (0*seg1 + 0*seg2 + 0*seg3 + 0*seg1 + 0*seg2 + 0*seg3)
What is the hit rate here? 0/0 is not zero. It is still undefined.
So the hit rate should be NaN NOT ZERO.

Kiminje:
I don't understand. So, do you mean derive_test shouldn't introduce divide by zero? 
But current derive_stats can't handle dbz. I know 0/0 is not zero, but at least it should not be plotted in the figure. 
Therefore, we have two cases that incur DNZ: it is never selected as a prediction or it is turned off so it never predicts
In my experiments, I frequently turn on/off TAGE components to evaluate how each component impacts other components. 
So, there are various configs in the same figure.

5surim:
> I don't understand. So, do you mean derive_test shouldn't introduce divide by zero? But current derive_stats can't handle dbz. I know 0/0 is not zero, but at least it should not be plotted in the figure. Therefore, we have two cases that incur DBZ: it is never selected as a prediction, or it is turned off, so it never predicts In my experiments, I frequently turn on/off TAGE components to evaluate how each component impacts other components. So, there are various configs in the same figure.
> 
> This is easily solved by adding a boolean to derive_stats to decide whether you allow DBZ or not. For example, as a default, let derive_stats don't pass DBZ because it's wrong. However, the user should able to select DBZ forcefully.
> 
> Otherwise, please let me know how to plot for my cases.
> 
> Sorry, I frequently type errors. DBZ is 'divided by zero'

Are you saying one config has a rate 0/0 and the other config has a rate non-zero/non-zero and plot them within a single plot?
Then for the value of the derived stat, keep the value NaN (0 is wrong in any case), and see how the matplotlib handles the NaN values. Plotting should handle this instead of directly changing the value in the dataframe.

5surim:
@Kiminje Can you update the title and description of this PR? I don't think the dbz/NaN is handled anymore in this PR.

5surim:
Let me close it since we have a plan.

PR #83, opened 2025-03-23
Add geekbench traces and simpoints (four frontend-bounded workloads)

5surim:


common/scripts/run_simpoint_trace.py
```diff
@@ -110,33 +120,63 @@ def trace_then_cluster(workload, suite, simpoint_home, bincmd, client_bincmd, si
...
+                        os.chmod(raw_path, 0o777)
+                        subprocess.run(f"cp {raw_path}/modules.log {bin_path}/modules.log", check=True, shell=True)
+                        subprocess.run(f"cp {raw_path}/modules.log {raw_path}/modules.log.bak", check=True, shell=True)
+                        subprocess.run(["python2", f"{simpoint_home}/scarab/utils/memtrace/portabilize_trace.py", f"{dr_path}"], capture_output=True, text=True, check=True)
+                        subprocess.run(f"cp {bin_path}/modules.log {raw_path}/modules.log", check=True, shell=True)
+                        raw2trace_cmd = f"{dynamorio_home}/tools/bin64/drraw2trace -jobs 40 -indir {raw_path} -chunk_instr_count {chunk_size}"
+                        process = subprocess.Popen("exec " + raw2trace_cmd, stdout=subprocess.PIPE, shell=True)
+                        raw2trace_processes.add(process)
 
         for p in raw2trace_processes:
```
moonseoppkim, line 181: What is the maximum number of sub-directories you plan to create?

I recently launched over 100 processes for iterative tracing mode (around 600 processes for raw2trace), which caused a significant load on our lab server.

If you have a large number of sub-directories, it might be better to split them into multiple stages to reduce the server load.

5surim: Good question.
We plan to refactor this tracing as well so that slurm workload manager controls the number of jobs (The same way we do for the simulation.)
For now, we don't limit the number of processes because the non-iterative tracing mode usually launches 1~10 processes.

moonseoppkim (approved):
LGTM

PR #87, opened 2025-03-26
Fix folder renaming timing for iterative tracing

moonseoppkim:
I made some changes to the iterative tracing sequence due to issues with running the simulation:

1. Renamed folders to follow the "Timestep_#" format before running raw2trace.
2. Updated the yabloc tracing folder.
3. Set a limit on the maximum process count during the raw2trace stage.

### DIFF
--- a/common/scripts/run_simpoint_trace.py
+++ b/common/scripts/run_simpoint_trace.py
@@ -403,7 +403,8 @@ def iterative(workload, suite, simpoint_home, bincmd, client_bincmd, simpoint_mo
                 except subprocess.TimeoutExpired as e:
                     print(f"[Timestep {i}] tracing attempt {attempt+1} timed out: {e}")
                     if attempt == max_retries - 1:
-                        raise e
+                        print(f"[Timestep {i}] tracing failed with error: {e}")
+                        break
                     else:
                         subdirs = [os.path.join(timestep_dir, d) for d in os.listdir(timestep_dir) 
                                    if os.path.isdir(os.path.join(timestep_dir, d))]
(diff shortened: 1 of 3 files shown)


5surim:
Hi @moonseoppkim 

I have some questions.
Why is the renaming required before running raw2trace? Is it due to modules.log?
Why have the weights of segments changed? Shouldn't the number of instructions for each segment be the same as before?

Thanks,
Surim

moonseoppkim:
Hi, @5surim 

Why is the renaming required before running raw2trace? Is it due to modules.log?
-> Yes, exactly.

It caused an issue at modules.log for loading libraries. Because, I renamed the whole folder structure after finishing raw2trace execution. 
I should've finished earlier before starting raw2trace.

```shell
/home/mkim293/simpoint_flow/yabloc/yabloc/traces_simp/Timestep_1/drmemtrace.call_yabloc_pose.00817.7089.dir/bin/libdrmemtrace.so
``` 

Why have the weights of segments changed? Shouldn't the number of instructions for each segment be the same as before?
->  Overall instruction counts are between 100,000,000 and 90,000,000, It is exactly same result as before.
```shell
100000000 0
100000000 1
100000000 2
100000000 3
90000000 4
100000000 5
``` 

Thanks.
Best regards

5surim:
> Hi, @5surim
> 
> Why is the renaming required before running raw2trace? Is it due to modules.log? -> Yes, exactly.
> 
> It caused an issue at modules.log for loading libraries. Because, I renamed the whole folder structure after finishing raw2trace execution. I should've finished earlier before starting raw2trace.
> 
> ```shell
> # previous error log
> Failed to load binaries: Failed to parse module file Check that module.log references the correct binary paths.Scarab started at Tue Mar 25 21:53:28 2025
> 
> # current success log
> /home/mkim293/simpoint_flow/yabloc/yabloc/traces_simp/Timestep_1/drmemtrace.call_yabloc_pose.00817.7089.dir/bin/libdrmemtrace.so
> ```
> 
> Why have the weights of segments changed? Shouldn't the number of instructions for each segment be the same as before? -> Overall instruction counts are between 100,000,000 and 90,000,000, It is exactly same result as before. I assumed a ratio between two different instruction numbers caused some changes to the weight file.
> 
> ```shell
> 100000000 0
> 100000000 1
> 100000000 2
> 100000000 3
> 90000000 4
> 100000000 5
> ```
> 
> Thanks. Best regards

If the instruction count of each timestep is exactly the same as before (e.g. Timestep_1: 90000000 then the new Timestep_1: 90000000), the weight of Timestep_1 should be identical, but your new workloads_db.json changed the weight. Could you elaborate why this happens?

moonseoppkim:
Unfortunately, I removed the previous tracing files already.
I can't compare directly between two tracing.

But, I guess the instruction count is extracting between two numbers(100,000,000 and 90,000,000) with individual timestep.
There was some differences in terms of the ratio of that two instruction numbers.

```shell
instruction count / cluster id
100000000 2
90000000 4
100000000 8
90000000 10
100000000 11
``` 

Do you think the entire instruction counts should be identical across all timesteps?

Thanks!

5surim:
> Unfortunately, I removed the previous tracing files already. I can't compare directly between two tracing.
> 
> But, I guess the iterative tracing is extracting two numbers(100,000,000 and 90,000,000) as instruction counts with individual timestep. There might have been some differences in terms of the ratio of those two instruction numbers.
> 
> ```shell
> instruction count / cluster id
> 100000000 2
> 90000000 4
> 100000000 8
> 90000000 10
> 100000000 11
> ```
> 
> Do you think the entire instruction counts should be identical across all timesteps?
> 
> Thanks!

It's okay to have a different number of instructions across all timesteps.
But for a specific timestep (say cluster_id 2), I believe it should always be 100000000 every time you re-run it.
And the total number of instructions (sum over all clusters) should be the same.
This will lead you to the weight (# of inst of cluster id / total # of instructions) = 0.0161287.. always. But it seems the weight changed.
If the workload binary and the input data are exactly the same, I believe it should be deterministic..?

But it's OK for now, I think these values are acceptable. Something happened we didn't figure out.

Is this patch ready-for-review?

moonseoppkim:
Yes, this patch is ready for review.

I am reviewing the entire tracing file again. There are no broken_trace files and everything is normal.
I'll keep an eye on it and let you know if I notice anything unusual.

5surim (approved):
LGTM!

PR #101, opened 2025-04-03
trim_type is written incorrect in finish_trace

5surim:
nine geekbench workloads got wrong trim_type when they were traced


moonseoppkim (approved):
LGTM

PR #105, opened 2025-04-04
Allow multiple interactive shells attached to a single container

5surim:
- only remove the container when the last shell exits

only affect --run
* you can run multiple `./run.sh --run <your_experiment>` commands (interactive shell) to log into a single container.


hlitz (approved):
LGTM

PR #109, opened 2025-04-08
Stat retreive fix

FlyingFish800:
Fixed error in retrieving stats that would occur when one config's name was a subset of another config, causing aggregation errors while plotting
Ex:
FDIP would also match Ideal_FDIP, and the weights of both runs would sum to 2.0

### DIFF
--- a/scarab_stats/scarab_stats.py
+++ b/scarab_stats/scarab_stats.py
@@ -91,7 +91,7 @@ class Experiment:
         if aggregation_level == "Workload":
             for c in config:
                 for w in workload:
-                    selected_simpoints = [col for col in self.data.columns if f"{c} {w}" in col]
+                    selected_simpoints = [col for col in self.data.columns if col.startswith(f"{c} {w}")]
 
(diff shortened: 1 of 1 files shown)


scarab_stats/scarab_stats.py
```diff
@@ -1575,7 +1579,24 @@ def load_simpoint_data(self, cluster_id, workload, config, experiment_name, arch
     args = parser.parse_args()
 
     da = stat_aggregator()
-    E = da.load_experiment_json(args.descriptor_name, True)
+    
+    # E = da.load_experiment_json(args.descriptor_name, True)
+    # E.to_csv("micro.csv")
+
+    E = da.load_experiment_csv("micro.csv")
```
hlitz, line 1586: I think we do not always want to load micro.csv
please check the other lines below as well.

hlitz (changes requested):

hlitz (approved):
LGTM

PR #110, opened 2025-04-08
[FIX] Only check available nodes for docker image

naomirehman1008:
The previous check for the docker image looked at all slurm nodes, which caused the script to fail if a node did not have docker installed. The fixed check only looks for the image on available nodes, which have already been checked for docker.


5surim (approved):
LGTM

PR #116, opened 2025-04-17
copy scarab binary only when the src/dest binaries differ or no dest binary eixsts

5surim:
This will allow running additional configurations within the same experiment directory
while it still checks if the scarab binary is busy when the binary is updated.

### DIFF
--- a/scripts/utilities.py
+++ b/scripts/utilities.py
@@ -218,12 +218,10 @@ def prepare_simulation(user, scarab_path, scarab_build, docker_home, experiment_
             except subprocess.CalledProcessError as e:
                 if e.returncode == 1 or e.returncode == 2:
                     info("scarab binaries differ or the destination binary does not exist. Will copy.", dbg_lvl)
-                    result = subprocess.run(["cp", scarab_bin, f"{experiment_dir}/scarab/src/scarab"],
-                                           capture_output=True,
-                                           text=True)
+                    result = subprocess.run(["cp", scarab_bin, dest_scarab_bin], capture_output=True, text=True)
                     if result.returncode != 0:
                         err(f"Failed to copy scarab binary: {result.stderr}", dbg_lvl)
-                        raise RuntimeError(f"Failed to copy scarab binary: {result.stderr}")
+                        raise RuntimeError(f"Failed to copy scarab binary. Existing binary is in use and differs from the new binary: {result.stderr}")
                 else:
                     raise e
         try:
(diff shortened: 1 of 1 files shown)


hlitz:
#96 introduced this check to prevent the following issue. When the binary is updated, the changes are not propagated because the copy is silently dropped. This caused unexpected behavior during debugging, if there is some older process still running using the older binary.

What is your intended use case? Add existing workloads to an existing experiment? Maybe we can do the following: diff the old and new scarab binary. If they are the same do not attempt the copy and start new simulations. If they are different, continue to fail if the old binary is still in use. So you are only allowed to skip the assertion if the scarab binary did not change.

5surim:
> #96 introduced this check to prevent the following issue. When the binary is updated, the changes are not propagated because the copy is silently dropped. This caused unexpected behavior during debugging, if there is some older process still running using the older binary.
> 
> What is your intended use case? Add existing workloads to an existing experiment? Maybe we can do the following: diff the old and new scarab binary. If they are the same do not attempt the copy and start new simulations. If they are different, continue to fail if the old binary is still in use. So you are only allowed to skip the assertion if the scarab binary did not change.

The use case I experienced in many cases was the following.
I first launched simulations for 'config1' and 'config2'.
I want to add more simulations for 'config3' and 'config4'. This should still use the same binary and recorded within the same 'experiment' named directory so that we can plot them all together.

Yes, I think I like the idea. Let me bring the patch for that.

5surim:
@hlitz This is ready for review

scripts/utilities.py
```diff
@@ -212,11 +212,20 @@ def prepare_simulation(user, scarab_path, scarab_build, docker_home, experiment_
...
+                result = subprocess.run(['diff', '-q', scarab_bin, dest_scarab_bin], capture_output=True, text=True, check=True)
+            except subprocess.CalledProcessError as e:
+                if e.returncode == 1 or e.returncode == 2:
+                    info("scarab binaries differ or the destination binary does not exist. Will copy.", dbg_lvl)
+                    result = subprocess.run(["cp", scarab_bin, f"{experiment_dir}/scarab/src/scarab"],
+                                           capture_output=True,
+                                           text=True)
+                    if result.returncode != 0:
+                        err(f"Failed to copy scarab binary: {result.stderr}", dbg_lvl)
+                        raise RuntimeError(f"Failed to copy scarab binary: {result.stderr}")
```
hlitz, line 226: say: Failed to copy scarab binary. Existing binary is in use and differs from the new binary.

hlitz (changes requested):

hlitz (approved):
LGTM

PR #121, opened 2025-04-17
Add an instruction for proper 'docker pull' and fix for it

5surim:
`docker pull` command became not working without proper permission setup after `Docker Image Manifest V2`.

Fix:
1. Instructions to correctly setup the requirement based on the [GitHub Doc `Working with the Container registry`](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry).
2. Check the output of `docker pull` command and handle it properly. Only build a new image when it is not available by `docker pull`


hlitz (approved):
LGTM

PR #122, opened 2025-04-18
Cutting scarab_stats execution time by 50%.

moonseoppkim:
Test env : 2800 times load_simpoint_data func calls

Before    : 40min (30min load_simpoint_data + 10min calculate_distribution_stats)
After       : 15min (5min load_simpoint_data + 10min calculate_distribution_stats)

Updating the pandas DataFrame column‑by‑column in add_simpoint() caused substantial slowdowns. 
So, I made changes to collect all new column data in a dictionary first and convert it to the DataFrame in one bulk operation at the final stage.

### DIFF
--- a/scarab_stats/scarab_stats.py
+++ b/scarab_stats/scarab_stats.py
@@ -43,6 +43,7 @@ class Experiment:
 
             # Enable write protect for all base stats
             self.data["write_protect"] = [True for _ in rows]
+            self._columns_dict = {}
 
     def has_group_data(self):
         return "groups" in self.data.columns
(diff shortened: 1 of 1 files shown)


5surim:
Thanks for the fix.
#80 also reduces the execution time.
@moonseoppkim @Kiminje, can we discuss this first before you deliver a fix? We probably don't want to spend each MH on the same issue.
@hlitz, can you review both, and can we discuss this execution time issue?

moonseoppkim:
Hi, @Kiminje,
I only updated the add_simpoint function, which seems like independent with your #80 PR.
If you have any questions, let me know!

hlitz:
Thanks @moonseoppkim I have a patch myself which is at 5min + 1 sec. I have been thinking about rewriting the entire architecture. Lets have a meeting on this first.

5surim:
Let me close it since we have a plan.

PR #126, opened 2025-04-21
Stat autocollection

FlyingFish800:
Runs a script to automatically collect stats for both the slurm and manual runners, puts a file named `collected_stats.csv` in the top level of the folder for the simulation. Ex: `/simulations/simulation_name/collected_stats.csv`. 

This will also work for any refactored scarab stats library, likely without any modifications to the runner script.


scripts/local_runner.py
```diff
@@ -257,6 +257,15 @@ def run_single_workload(workload, exp_cluster_id, sim_mode):
         for p in processes:
             p.wait()
 
+        # Generate stats csvs
+        print(f"Collecting stats...")
+        
+        # Import scarab stats for auto-generating csvs
+        subprocess.check_output(["python3", "stat_collector.py",
```
5surim, line 264: I can't find `stat_collector.py`. It seems you didn't add it to this PR.
Also, instead of running this within `run_simulation()` function of `local_runner.py` and `slurm_runner.py`, can you call this in `run_simulation.py`? I think this doesn't depend on which workload manager is used.

scripts/slurm_runner.py
```diff
@@ -439,6 +447,11 @@ def run_single_workload(workload, exp_cluster_id, sim_mode):
 
         finish_simulation(user, docker_home)
 
+        print(slurm_ids)
+        collect_stats_cmd = f"sbatch --dependency=afterok:{','.join(slurm_ids)} scripts/slurm_collect_stats.sh {descriptor_path} {descriptor_data['root_dir']} {experiment_name}"
```
5surim, line 451: Does this guarantee that it waits for all the slurm jobs finished?

5surim: I also don't find `slurm_collect_stats.sh`.
I would make it wait for all the simulations done, then collect stats after that. This will allow you to do the same collection logic for both `local_runner` and `slurm_runner`

5surim (changes requested):
Hi @FlyingFish800 ,

thanks for the patch.
I think there are missing files you didn't add to this PR.
Please check out the comments.

Surim

scripts/slurm_runner.py
FlyingFish800: I think I could also move the run commands to finish_simulations, and use an optional parameter to prepend the sbatch command to the normal collection logic, as follows:
```
# For slurm, add slurm dependencies
if slurm_ids != None:
    collect_stats_cmd = f"sbatch --dependency=afterok:{','.join(slurm_ids)}" + collect_stats_cmd

# Rest of logic is identical
```
This would preserve the asynchronous nature of the slurm runner, if that is desirable. Is adding the wait logic preferable to this method?

5surim: I like the asynchronous nature of the Slurm runner, but does it collect stats if it doesn't wait for all the jobs to finish? I guess this command will always fail to collect stats because it tries to collect the stats right after launching the slurm jobs.
Instead of waiting for it, we can add a callback function that is triggered when all the simulations from the experiment are successfully finished.

5surim: Maybe `--dependency=afterok ` is the one (callback)?

5surim: Also, instead of using shell script, could you use python with a proper exception handling?
My experience is that if it fails somewhere within a bash shell script, it is hard to figure out where it failed. I would always throw the exceptions within python script.

FlyingFish800: Ok, I  can try to cut out the shell script.

With `--dependency=afterok:<ids>`, that job will only run if all the dependent jobs exit successfully, so all required data will be present. It does wait for other jobs to finish, but that is handled by the slurm scheduler.

FlyingFish800:
Since the last review I added the missing files, cut out the shell script, and run collection in finish_simulation(). The asynchronous slurm behaviour is preserved, and the log is written to a specially marked log file with the others in the case of using slurm.

5surim (approved):
LGTM.
Please extend the stat_collector when you extend --status command.

PR #127, opened 2025-04-22
Use a hierarchical trace directories for simulation

5surim:
- read traces from hierarchical directories (suite/subsuite/workload)
- unify trace path for different trace modes (remove trim_type)
- warmup field introduced to support different warmup (# of inst) for each workload
- stats result also in hierarchical directories (config/suite/subsuite/workload/cluster_id)
- remove `suite_db.json` and add suite/subsuite to `workloads_db.json` (json field accessed by suite->subsuite->workload)


5surim:
When this patch gets merged, we should release another top_simpoint.tar.gz to change the workload name (not 500.perlbench_r, instead, spec2017/rate_int/perlbench)

5surim:
@hlitz I think this is ready for review.
@FlyingFish800 Please work on stats library rewriting on top of this.

5surim:
I just found the fixes missing in `run_trace.py`. Let me bring them.

hlitz (approved):
LGTM

PR #131, opened 2025-04-28
Make the container name unique with suite, subsuite names

5surim:
Make the container name unique with suite, subsuite names to avoid job dependency in slurm


hlitz (approved):
LGTM

Issue #134, opened 2025-05-01
run.sh --status my.json


hlitz:
Show the status of an experiment as follows:

| Configuration | Completed | Failed | Running | Pending | Total |
|---------------|--------|---------|---------|-------|-------|
| baseline      | 5      | 1       | 5       | 2     | 13    |
| config_1      | 5      | 1       | 5       | 2     | 13    |

Failed runs: (show up to 5 max)
/home/user/../sim.log

PR #135, opened 2025-05-04
scarab build at the selected slrum node

moonseoppkim:
There was a bug in the Scarab build process in slurm_runner.py.
So, I added target_node parameter for all docker related subprocess.run function calls.

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -211,7 +211,7 @@ run_scarab () {
   taskPids=()
   start=`date +%s`
 
-  cmd="python3 ${INFRA_ROOT}/scripts/run_simulation.py -dbg 1 -d ${INFRA_ROOT}/json/${SIMULATION}.json"
+  cmd="python3 ${INFRA_ROOT}/scripts/run_simulation.py -dbg 3 -d ${INFRA_ROOT}/json/${SIMULATION}.json"
   eval $cmd &
   taskPids+=($!)
 
--- a/run.sh
+++ b/run.sh
@@ -100,9 +100,11 @@ run () {
 
   # Key to extract
   key="descriptor_type"
+  workload_manager_key="workload_manager"
 
   # Extract the value using fq
   value=$(jq -r ".$key" "$json_file")
+  workload_manager=$(jq -r ".$workload_manager_key" "$json_file")
 
   if [ $value == "simulation" ]; then
     workload_db_json_file="${INFRA_ROOT}/workloads/workloads_db.json"
@@ -122,8 +124,10 @@ run () {
 
     # check if the Docker image '$APP_GROUPNAME:$GIT_HASH' exists
     if ! docker images --format "{{.Repository}}:{{.Tag}}" | grep -q "$APP_GROUPNAME:$GIT_HASH"; then
-      echo "The image with the current Git commit hash does not exist! Build the image first by using '-b'."
-      exit 1
+      if [ $workload_manager == "manual" ]; then
+        echo "The image with the current Git commit hash does not exist! Build the image first by using '-b'."
+        exit 1
+      fi
     fi
 
 
--- a/scripts/run_simulation.py
+++ b/scripts/run_simulation.py
@@ -7,6 +7,7 @@ import subprocess
 import argparse
 import os
 import docker
+import random
 
 from utilities import (
         info,
@@ -18,7 +19,11 @@ from utilities import (
         get_image_name,
         validate_simulation,
         is_container_running,
-        count_interactive_shells
+        count_interactive_shells,
+        check_available_nodes,
+        check_docker_image,
+        check_docker_container_running,
+        run_on_node
         )
 import slurm_runner
 import local_runner
@@ -94,6 +99,7 @@ def verify_descriptor(descriptor_data, workloads_data, open_shell = False, dbg_l
 def open_interactive_shell(user, descriptor_data, workloads_data, infra_dir, dbg_lvl = 1):
     experiment_name = descriptor_data["experiment"]
     scarab_path = descriptor_data["scarab_path"]
+    workload_manager = descriptor_data["workload_manager"]
     try:
         # Get user for commands
         user = subprocess.check_output("whoami").decode('utf-8')[:-1]
@@ -127,6 +133,15 @@ def open_interactive_shell(user, descriptor_data, workloads_data, infra_dir, dbg
             if entry not in f.read():
                 f.write(f"\n{entry}\n")
 
+        target_node = None
+        if workload_manager == "slurm":
+            available_slurm_nodes = check_docker_image(check_available_nodes(dbg_lvl)[0], docker_prefix, githash, dbg_lvl)
+            if len(available_slurm_nodes) == 0:
+                err("The image with the current Git commit hash does not exist! Build the image first by using '-b'.", dbg_lvl)
+                exit(1)
+            target_node = random.choice(available_slurm_nodes)
+            print("generating interactive shell, target_node: ", target_node)
+
         # Generate commands for executing in users docker and sbatching to nodes with containers
         scarab_githash = prepare_simulation(user,
                                             scarab_path,
@@ -138,7 +153,8 @@ def open_interactive_shell(user, descriptor_data, workloads_data, infra_dir, dbg
                                             githash,
                                             infra_dir,
                                             True,
-                                            dbg_lvl)
+                                            dbg_lvl,
+                                            target_node)
         workload = descriptor_data['simulations'][0]['workload']
         mode = descriptor_data['simulations'][0]['simulation_type']
 
@@ -147,10 +163,11 @@ def open_interactive_shell(user, descriptor_data, workloads_data, infra_dir, dbg
         docker_home = descriptor_data["root_dir"]
         try:
             # If the container is already running, log into it by openning another interactive shell
-            if is_container_running(docker_container_name, dbg_lvl):
-                subprocess.run(["docker", "exec", "--privileged", "-it", f"--user={user}", f"--workdir=/home/{user}", docker_container_name, "/bin/bash"])
+            if is_container_running(docker_container_name, dbg_lvl, target_node):
+                run_on_node((["--pty"] if workload_manager == "slurm" else []) + 
+                             ["docker", "exec", "--privileged", "-it", f"--user={user}", f"--workdir=/home/{user}", docker_container_name, "/bin/bash"], node=target_node)
             else:
-                subprocess.run(["docker", "run",
+                run_on_node(["docker", "run",
                                 "-e", f"user_id={local_uid}",
                                 "-e", f"group_id={local_gid}",
                                 "-e", f"username={user}",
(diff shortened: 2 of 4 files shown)


5surim:
> There was a bug in the Scarab build process in slurm_runner.py. So, I added target_node parameter for all docker related subprocess.run function calls.

Can you describe what the bug exactly is and the root cause of it?

moonseoppkim:
> > There was a bug in the Scarab build process in slurm_runner.py. So, I added target_node parameter for all docker related subprocess.run function calls.
> 
> Can you describe what the bug exactly is and the root cause of it?


If we set a scarab_build option ("opt/dgb/etc"), we should run the command on the selected slurm node where the container was created.

But, the original source code executes subprocess.call on bohr1(local).

moonseoppkim:
> > > There was a bug in the Scarab build process in slurm_runner.py. So, I added target_node parameter for all docker related subprocess.run function calls.
> > 
> > 
> > Can you describe what the bug exactly is and the root cause of it?
> 
> If we set a scarab_build option ("opt/dgb/etc"), we should run the scarab build command on the selected slurm node where the container was created.
> 
> But, the original source code executes subprocess.call on bohr1(local).

moonseoppkim:
> > > There was a bug in the Scarab build process in slurm_runner.py. So, I added target_node parameter for all docker related subprocess.run function calls.
> > 
> > 
> > Can you describe what the bug exactly is and the root cause of it?
> 
> If we set a scarab_build option ("opt/dgb/etc"), we should run the scarab build command on the selected slurm node where the container was created.
> 
> But, the original source code executes subprocess.call on bohr1(local).

5surim:
`prepare_simulation` has been called only once before launching many simulation jobs on local, which is now bohr1 for everyone. @FlyingFish800 Is the reason we don't find the image on bohr1 that bohr1 is now the control node and not the worker node anymore?
```
sinfo -N
NODELIST   NODES PARTITION STATE
bohr2          1   LocalQ* idle
bohr3          1   LocalQ* idle
bohr4          1   LocalQ* drain
bohr5          1   LocalQ* idle
ohm            1   LocalQ* mix
```

FlyingFish800:
@5surim yes, I believe that is correct. Bohr1 cannot run any slurm jobs anymore, only dispatch them to other nodes, so the image won’t be built into Bohr1

moonseoppkim:
> @5surim yes, I believe that is correct. Bohr1 cannot run any slurm jobs anymore, only dispatch them to other nodes, so the image won’t be built on Bohr1

Hi! Alex

Could you answer my question above?

5surim:
> @5surim yes, I believe that is correct. Bohr1 cannot run any slurm jobs anymore, only dispatch them to other nodes, so the image won’t be built on Bohr1

I think we still need the image on bohr1 for the purpose of debugging or any interactive shell.
Can we still build the image also locally, although we don't allow the simulations via slurm on bohr1?
So, only include bohr1 into the node list for building an image.

@moonseoppkim I think we still want to build scarab locally so that we can quickly open the debug shell when we see any build error.

moonseoppkim:
we can still use a "manual" option for debugging purpose on bohr1 instead of "slurm".

The problem was that the current slurm_runner couldn't handle the scarab_build logic properly.
* prepare docker images for scarab build       [bohr2 - ohm]
* generating scarab build container                 [bohr1]

5surim:
> we can still use a "manual" option for debugging purpose on bohr1 instead of "slurm".
> 
> The problem was that the current slurm_runner couldn't handle the scarab_build logic properly.
> 
> * prepare docker images for scarab build       [bohr2 - ohm]
> * generating scarab build container                 [bohr1]

Yes, I understand what is happening now.
The common use case is that
1) run simulation via slurm
2) scarab build fails
3) open an interactive shell without modifying JSON

I think we should still build the image locally before launching the simulation, because it will fail before launching the simulation, which is an early error return.

moonseoppkim:
> > we can still use a "manual" option for debugging purpose on bohr1 instead of "slurm".
> > The problem was that the current slurm_runner couldn't handle the scarab_build logic properly.
> > 
> > * prepare docker images for scarab build       [bohr2 - ohm]
> > * generating scarab build container                 [bohr1]
> 
> Yes, I understand what is happening now. One of the common use cases is the following
> 
> 1. run simulation via slurm
> 2. scarab build fails
> 3. open an interactive shell without modifying JSON
> 
> I think we should still build the image locally before launching the simulation, because if it fails before launching the simulation, it is also an early error return.

I got it, 

we need one more information to handle this situation.
1. we can have multiple containers which have same name across all servers.
2. we should specify which server's container was the latest one.

Therefore, we need to know the latest fail history to pick the latest "server/container name" combination in order to open an interactive shell. The similar issue has been opened with #134.

moonseoppkim:
> > > > we can still use a "manual" option for debugging purpose on bohr1 instead of "slurm".
> > > > The problem was that the current slurm_runner couldn't handle the scarab_build logic properly.
> > > > 
> > > > * prepare docker images for scarab build       [bohr2 - ohm]
> > > > * generating scarab build container                 [bohr1]
> > > 
> > > 
> > > Yes, I understand what is happening now. One of the common use cases is the following
> > > 
> > > 1. run simulation via slurm
> > > 2. scarab build fails
> > > 3. open an interactive shell without modifying JSON
> > > 
> > > I think we should still build the image locally before launching the simulation, because if it fails before launching the simulation, it is also an early error return.
> > 
> > 
> > I got it,
> > we need one more information to handle this situation.
> > 
> > 1. we can have multiple containers which have same name across all servers.
> > 2. we should specify which server's container was the latest one. (or put additional parameter to pick a specific server)
> > 
> > Therefore, we need to know the fail history to pick the latest "server/container name" combination in order to open an interactive shell. The similar issue has been opened with #134.
> 
> 1. I believe this is only possible when the same JSON is used for re-run. Are you assuming the first run has some failures but without killing the old ones, you run the experiment again? or are you still seeing the same named containers even without any failure?


I thought it was possible, but it's a very edge case in general.
I can add more codes for opening a interactive shell based on the container name.

* sbatch workflow
1. run simulation with (./run.sh --simulation)
2. slurm node check available nodes like below,
```python
NODELIST   NODES PARTITION STATE 
bohr2          1   LocalQ* idle  
bohr3          1   LocalQ* idle  
bohr4          1   LocalQ* drain 
bohr5          1   LocalQ* idle  
ohm            1   LocalQ* idle 
``` 
3. run sbatch command from slurm_runner.py
```python
"sbatch --exclude {','.join(excludes)} -c 1 -o {experiment_dir}/logs/job_%j.out "
``` 
4. sbatch assigns jobs sequentially, starting with bohr2 to ohm only after the previous node’s cores are filled.
```python
/etc/slurm/slurm.conf

SelectType=select/cons_tres -> keeping the number of activated core to minimum
SelectTypeParameters=CR_Core
``` 
5. A simulation which has same container name will run on the bohr2 again and again and got fail to run as the duplicated container issue.
6. Only the first "docker run command" was valid.

5surim:
> > > > > we can still use a "manual" option for debugging purpose on bohr1 instead of "slurm".
> > > > > The problem was that the current slurm_runner couldn't handle the scarab_build logic properly.
> > > > > 
> > > > > * prepare docker images for scarab build       [bohr2 - ohm]
> > > > > * generating scarab build container                 [bohr1]
> > > > 
> > > > 
> > > > Yes, I understand what is happening now. One of the common use cases is the following
> > > > 
> > > > 1. run simulation via slurm
> > > > 2. scarab build fails
> > > > 3. open an interactive shell without modifying JSON
> > > > 
> > > > I think we should still build the image locally before launching the simulation, because if it fails before launching the simulation, it is also an early error return.
> > > 
> > > 
> > > I got it,
> > > we need one more information to handle this situation.
> > > 
> > > 1. we can have multiple containers which have same name across all servers.
> > > 2. we should specify which server's container was the latest one. (or put additional parameter to pick a specific server)
> > > 
> > > Therefore, we need to know the fail history to pick the latest "server/container name" combination in order to open an interactive shell. The similar issue has been opened with #134.
> > 
> > 
> > 
> > 1. I believe this is only possible when the same JSON is used for re-run. Are you assuming the first run has some failures but without killing the old ones, you run the experiment again? or are you still seeing the same named containers even without any failure?
> 
> I thought it was possible, but it's a very edge case in general. I can add more codes for opening a interactive shell based on the container name.
> 
> * sbatch workflow
> 
> 1. run simulation with (./run.sh --simulation)
> 2. slurm node check available nodes like below,
> 
> ```python
> NODELIST   NODES PARTITION STATE 
> bohr2          1   LocalQ* idle  
> bohr3          1   LocalQ* idle  
> bohr4          1   LocalQ* drain 
> bohr5          1   LocalQ* idle  
> ohm            1   LocalQ* idle 
> ```
> 
> 3. run sbatch command from slurm_runner.py
>    ** I thought the sbatch's assigning policy was random out of all idle nodelists, but it wasn't true.
> 
> ```python
> "sbatch --exclude {','.join(excludes)} -c 1 -o {experiment_dir}/logs/job_%j.out "
> ```
> 
> 5. sbatch assigns jobs sequentially, starting with bohr2 to ohm only after the previous node’s cores are filled.
> 
> ```python
> /etc/slurm/slurm.conf
> 
> SelectType=select/cons_tres -> keeping the number of activated core to minimum
> SelectTypeParameters=CR_Core
> ```
> 
> 6. A simulation which has same container name will run on the bohr2 again and again and got fail to run as the duplicated container issue.
> 7. Only the first "docker run command" was valid.

1~5 are exactly aligned with my understanding.
If the simulation never fails or any failure has been addressed in a clean way, 6 should not happen.
The container name is determined by `f"{docker_prefix}_{suite}_{subsuite}_{workload}_{experiment_name}_{config_key.replace("/", "-")}_{cluster_id}_{sim_mode}_{user}"`
So it should be unique if the previous run was successful. I think the old ones have not been cleaned up properly (docker container was not properly stopped and removed), and re-running the same json finds the same name container. Let me know if it is not the case, then it is another bug.
If that is the case, we have an issue regarding the failure handling. We can address this when we address the issue.

This PR addresses two different issues.
One is for checking if the docker image exists, and the other is regarding scarab build.
I think the exception handling (image_exist call on the node) is a good bug fix. Could you only deliver it first?
Thanks!

moonseoppkim:
In steps 6 and 7, I explained why no duplicate containers were created on the other servers, 
even though I ran the same command several times before the previous task had finished.

Like you mentioned, current PR is addressing 2 issues.
1. image_exist
2. scarab build subprocess run on the specific target node

The both changes are required to solve a bug related to the "scarab binary build logic". (not only for image_exist)

5surim:
> Sorry for confusing comments, in steps 6 and 7 which I explained why no duplicate containers were created on the other servers, even though I ran the same command several times before the previous task had finished.
> 
> * I tried multiple simulation runs intentionally for testing purpose, not introducing another issue.
> 
> Like you mentioned, current PR is addressing 2 issues.
> 
> 1. image_exist
> 2. scarab build subprocess run on the specific target node
> 
> The both changes are required to solve a bug related to the "scarab binary build logic". (not only for image_exist)

1 resolves a bug regarding the exception handling. If the image does not exist even after the build command `run.sh -b <image_name>` on the node, we should check if the image exists or not on that node.

2 resolves the bug you mentioned, but we don't want to resolve the bug by running the build command on the node. We still want to build scarab locally.

If you want to resolve the bugs together, can you make it build the image locally as well and build scarab locally instead of on the slurm node?

PR #138, opened 2025-05-06
Fix a bug on the image_exist exception handler

moonseoppkim:
image_exist function bug fix

The image_exist implementation didn't cover the slurm use-cases.
If "docker pull fail" occurs, we should run the image_check function to the target_node, not local.


5surim (approved):
LGTM! Thanks for fixing the bug.

PR #143, opened 2025-05-07
Status option update

FlyingFish800:
This includes the updates to the slurm job status. Per config status printing was added as was requested, and the per-node running simulation counts was retained.


scripts/slurm_runner.py
```diff
@@ -212,47 +254,302 @@ def launch_docker(infra_dir, docker_home, available_nodes, node=None, dbg_lvl=1)
...
+
+    # print(f"\033[92mSuccessfully Completed Jobs: {len(all_jobs) - len(not_complete) - len(error_runs)}\033[0m")
+    
+    data = {"Configuration":[],"Completed":[],"Failed":[],"Running":[],"Pending":[],"Total":[]}
+    for conf in confs:
+        data["Configuration"].append(conf)
+        data["Completed"].append(completed[conf])
+        data["Failed"].append(failed[conf])
+        data["Running"].append(running[conf])
+        data["Pending"].append(pending[conf])
```
hlitz, line 530: Can you add a column "Nonexisting" for all runs that have never been run (dir does not exist)
Can you separate Failed into "Failed" and "Slurm-error"

scripts/slurm_runner.py
```diff
@@ -212,47 +254,311 @@ def launch_docker(infra_dir, docker_home, available_nodes, node=None, dbg_lvl=1)
...
+    
+    data = {"Configuration":[],"Completed":[],"Failed":[],"Failed - Slurm":[],"Running":[],"Pending":[],"Non-existant":[],"Total":[]}
+    for conf in confs:  
+        data["Configuration"].append(conf)
+        data["Completed"].append(completed[conf])
+        data["Failed"].append(failed[conf])
+        data["Failed - Slurm"].append(slurm_failed[conf])
+        data["Running"].append(running[conf])
+        data["Pending"].append(pending[conf])
+        total_found = completed[conf] + failed[conf] + running[conf] + pending[conf]
```
hlitz, line 538: I think we should add slurm_failed and non-existant also to total_found

hlitz (changes requested):

scripts/slurm_runner.py
FlyingFish800: I added slurm_failed to the total_found. I changed the 'Total' row to print the calculated total number of simpoints that should exist per simpoint. This accounts for every possible state a job could be in

hlitz (approved):
LGTM

PR #146, opened 2025-05-08
Fix ModuleNotFoundError which results in workflow failure

yzhao321:
# Intro
Previous commit add `__init__` makes workload become module, but it is still not found.
This commit fixes ModuleNotFoundError by replacing relative sys.path.append("../") with an absolute path insertion.

# Eval
Works in an env without conda

### DIFF
--- a/scripts/utilities.py
+++ b/scripts/utilities.py
@@ -9,11 +9,6 @@ import re
 import docker
 import importlib
 import sys
-
-# Add the project root to sys.path
-project_root = os.path.abspath(os.path.join(os.path.dirname(__file__), ".."))
-sys.path.insert(0, project_root)
-
 import workloads.extract_top_simpoints as extract_top_simpoints
 importlib.reload(extract_top_simpoints)
 
--- a/scripts/utilities.py
+++ b/scripts/utilities.py
@@ -10,15 +10,12 @@ import docker
 import importlib
 import sys
 
-# Temporarily add the project root to sys.path
+# Add the project root to sys.path
 project_root = os.path.abspath(os.path.join(os.path.dirname(__file__), ".."))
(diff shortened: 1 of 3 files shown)


scripts/utilities.py
```diff
@@ -9,10 +9,16 @@
...
+
+# Temporarily add the project root to sys.path
+project_root = os.path.abspath(os.path.join(os.path.dirname(__file__), ".."))
+sys.path.insert(0, project_root)
+
 import workloads.extract_top_simpoints as extract_top_simpoints
 importlib.reload(extract_top_simpoints)
 
+# Restore sys.path to avoid polluting the global import space
+sys.path.pop(0)
```
5surim, line 21: How does this adding the project root to sys.path pollut the global import space?
Did you see any error regarding this?
Otherwise, I would just add it to the path not only for this `utilities.py` but for the entrypoint python script `run_simulation.py`.

5surim (commented):
@yzhao321 Thanks for the patch. Have you tested this patch if it works or not? If so, could you share how you tested it?

5surim:
@yzhao321 Can you test the new patch?

yzhao321:
@5surim 
Works on an env without conda <-- I tested it in my local side without conda
Tested after the second commit.

5surim:
@yzhao321 I don't see the error you shared (workloads module not found) with the third commit on my end (without conda on bohr3), instead, I saw a missing import `sys`, so I added.

5surim:
I would just add it to PYTHONPATH.

5surim (approved):
LGTM

5surim:
Even without the fix, on my end without conda on bohr3, I don't see the error.

PR #149, opened 2025-05-09
Fix import issue

5surim:
I will bring a better fix in a later PR.
Let's merge YY's fix for now.

@yzhao321, sorry for reopening PR. I had to force-push by reverting my commits. Please add your description of this PR here if needed.

### DIFF
--- a/scripts/utilities.py
+++ b/scripts/utilities.py
@@ -10,15 +10,12 @@ import docker
 import importlib
 import sys
 
-# Temporarily add the project root to sys.path
+# Add the project root to sys.path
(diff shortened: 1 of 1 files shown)


5surim:
@5surim : don't see the error even without conda on bohr3.
@yzhao321 : see the error on bohr3 without conda, but the fix removed the error.
@yql5510718 : even with the current fix, the error still exists.

I think something weird is happening. Let's dig into this.

yql5510718:
Update: With this fix, infra works for me on bohr3 without conda and bohr1 with conda.

yql5510718 (approved):
LGTM

PR #153, opened 2025-05-14
Added overwrite protection, re-run non-existant or failed

FlyingFish800:
Added logic to prevent overwriting of existing runs, and allow overwriting of errored/non-existant runs. Implemented identically in manual and slurm, logic is implemented in utilities.py

### DIFF
--- a/scripts/slurm_runner.py
+++ b/scripts/slurm_runner.py
@@ -656,7 +656,7 @@ def run_simulation(user, descriptor_data, workloads_data, infra_dir, descriptor_
 
                 for cluster_id, weight in simpoints.items():
                     info(f"cluster_id: {cluster_id}, weight: {weight}", dbg_lvl)
-                        
+
                     docker_container_name = f"{docker_prefix}_{suite}_{subsuite}_{workload}_{experiment_name}_{config_key.replace("/", "-")}_{cluster_id}_{sim_mode}_{user}"
 
                     # TODO: Notification when a run fails, point to output file and command that caused failure
--- a/scripts/slurm_runner.py
+++ b/scripts/slurm_runner.py
@@ -656,10 +656,6 @@ def run_simulation(user, descriptor_data, workloads_data, infra_dir, descriptor_
 
                 for cluster_id, weight in simpoints.items():
                     info(f"cluster_id: {cluster_id}, weight: {weight}", dbg_lvl)
-
-                    if check_can_skip(descriptor_data, config_key, suite, subsuite, workload, cluster_id, dbg_lvl):
-                        info(f"Skipping {workload} with config {config_key} and cluster id {cluster_id}", dbg_lvl)
-                        continue
                         
                     docker_container_name = f"{docker_prefix}_{suite}_{subsuite}_{workload}_{experiment_name}_{config_key.replace("/", "-")}_{cluster_id}_{sim_mode}_{user}"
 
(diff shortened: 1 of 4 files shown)


scripts/utilities.py
```diff
@@ -790,6 +797,102 @@ def image_exist(image_tag, node=None):
...
+        if os.path.exists(filename):
+            # Run script has generated run file, it will be run shortly.
+            info(f"Run script for {config_key} for workload {workload} exists. Other script will run it.", debug_lvl)
+            return True
+
+        # If using slurm, check queue too
+        if not slurm_queue is None:
+            # Check each entry
+            for entry in slurm_queue:
+                # Check for following identifier. Should be of form <docker_prefix>_...as below..._<trace_method>_<user>
```
5surim, line 885: I guess `_<trace_method>_<user>` is a typo?

scripts/utilities.py
```diff
@@ -790,6 +797,102 @@ def image_exist(image_tag, node=None):
...
+    experiment_dir =  f"{descriptor_data['root_dir']}/simulations/{descriptor_data['experiment']}/"
+    experiment_dir += f"{config_key}/{suite}/{subsuite}/{workload}/{exp_cluster_id}"
+
+    for file in os.listdir(experiment_dir):
+        try:
+            os.remove(os.path.join(experiment_dir, file))
+        except Exception as e:
+            err(f"Error removing file {file}: {e}", 1)
+
+    os.rmdir(experiment_dir)
```
5surim, line 840: Do we have a reason to remove each file instead of removing the directory at once? (`rm -rf`)
Also, os.rmdir() will throw an exception if the directory is not empty. It might be good to also catch the exception and raise it.

scripts/utilities.py
```diff
@@ -790,6 +797,102 @@ def image_exist(image_tag, node=None):
...
+    log_dir =  f"{descriptor_data['root_dir']}/simulations/{descriptor_data['experiment']}/logs/"
+    log_files = os.listdir(log_dir)
+    for file in log_files:
+        with open(os.path.join(log_dir, file), 'r') as f:
+            lines = f.readlines()
+
+            # Logfile will have {config} {suite}/{subsuite}/{workload} {simpoint} as header
+            header = f"{config_key} {suite}/{subsuite}/{workload} {exp_cluster_id}"
+            if header in lines:
+                lines.remove(header)
```
5surim, line 852: Why do you need to remove the header line? It seems you will delete the file anyway?

scripts/slurm_runner.py
```diff
@@ -655,7 +656,7 @@ def run_single_workload(suite, subsuite, workload, exp_cluster_id, sim_mode, all
 
                 for cluster_id, weight in simpoints.items():
                     info(f"cluster_id: {cluster_id}, weight: {weight}", dbg_lvl)
-
+
```
5surim, line 659: Please remove the white spaces.

5surim (changes requested):
Hi @FlyingFish800 ,

Thanks for the patch.
Could you check the comments?

Surim

scripts/utilities.py
FlyingFish800: It is redundant, I will remove it

scripts/utilities.py
FlyingFish800: Ah yes, I believe the docker containers are named `..._<sim_mode>_<user>`. When I was printing the `slurm_queue` list to figure out the sub-string to compare, I saw pt/memtrace and assumed it was the trace method. I will change the comment to have the same form as it is generated in `slurm_runner`

5surim (approved):
LGTM

PR #158, opened 2025-05-26
Only build if necessary

5surim:
Track the latest pre-built image's tag within a file `last_built_tag.txt` and always pull the correct image.
From now on, the latest tag is not the same as the latest git hash of the main branch.
Prepare the images as the preparation step just before launching the jobs within `prepare_simulation` and clean up the images in `finish_simulation`.

Update the workflow to build the image and update the tag only when the changes are detected in the image itself.


FlyingFish800 (changes requested):
I found two small errors, but everything else seems good and worked when testing

5surim:
https://github.com/litz-lab/scarab_ll/pull/322
should be merged together.

FlyingFish800 (approved):
LGTM!

PR #165, opened 2025-06-11
bug fix for the extract_top_simpoint func

moonseoppkim:
There was no matching function for extract_top_simpoints.modify_simpoints_in_place in extract_top_simpoints.py.
And, extract_workloads_with_simpoints seems like the core entry func for this script. So, I made this change.


5surim (approved):
LGTM.
Thanks for finding this. I missed the fix here when I changed the API name.

PR #178, opened 2025-06-29
Add power analysis utility for McPAT

yzhao321:
This commit add power analysis feature for infra

### DIFF
--- a/scarab_stats/power/McPATConverter.py
+++ b/scarab_stats/power/McPATConverter.py
@@ -1,0 +1,409 @@
+import os
+import json
+import math
+import xml.etree.ElementTree as ET
+import xml.dom.minidom as minidom
+from collections import OrderedDict
+
+
+class McPATConverter:
+    """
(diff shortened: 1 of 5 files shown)


5surim (changes requested):
Hi @yzhao321 ,

Thanks for the patch.
A couple of points:

We shouldn’t keep the newly generated simulation outputs in the infra repo. These should be automatically updated by reading the JSON file and retrieving the correct output path from it.

mcpat is an executable, so we shouldn’t commit the binary itself. Instead, we should ensure the correct version can be built when needed, similar to how we handle the fingerprint. You can include it in common/Dockerfile.common if we want it available across all containers.

Also, please provide an example of the generated power stats (a sample output) by following the AI-generated README instructions so we can confirm that everything works as expected.

Also, please make sure this workflow works in the scarabinfra conda environment, as it may not run correctly for others in different environments.

Thanks,
Surim

PR #192, opened 2025-08-17
Simpoint3.2 source integration

moonseoppkim:
Hi, Surim

This is the Simpoint3.2 integration commit.
* from https://github.com/kofyou/SimPoint.3.2.git

Thanks.


5surim:
Hi @moonseoppkim ,

It turns out that we already have SimPoint fork in LitzLab (https://github.com/litz-lab/SimPoint.3.2).
Can you close this PR?
Instead of cloning Mingsheng's, let's clone the one in LitzLab.
Can you first compare two different repo and figure out if Mingsheng's is a fork of the one in LitzLab or they are both forks of the original repo?
It will be great to have a PR to SimPoint (LitzLab) from Mingsheng's (authored by Mingsheng committed by you) if there is difference between the two.

Thanks,
Surim

PR #200, opened 2025-10-08
Config versioning

FlyingFish800:
Added ability to define scarab githash for each config
- Configs are now dictionaries, `scarab_params` entry is parameters and `scarab_githash` entry specifies git hash to checkout
- Checks new local scarab build cache (`scarab-infra/scarab_builds`) for binaries before building
- Checks out any missing git commit and builds binary for it. 'current' uses currently checked out commit
- Returns scarab repo to the commit it was originally checked out as after building binaries
- Changes to scarab **must** be stashed/committed before running with scarab-infra
- Copies of scarab binary are copied from cache folder into experiment directory for reproducibility
- Race conditions are avoided by file based lock. Other build processes will error and crash if repo is locked

Example of a new config entry:
```
"baseline": {
      "scarab_params": "--bp_mech tage64k --fdip_enable 1 --btb_entries 8192 --perfect_crs 1 --wp_collect_stats 1 --mem_req_buffer_entries 64 --ramulator_readq_entries 64 --fe_ftq_block_num 32",
      "scarab_githash": "current"
}
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+workloads/spec2017/cpu2017-1_0_5.iso
+scarab_builds/*
--- a/common/scripts/run_exec_single_simpoint.sh
+++ b/common/scripts/run_exec_single_simpoint.sh
@@ -16,6 +16,7 @@ SEGMENT_ID="$8"
 ENVVAR="$9"
 BINCMD="${10}"
 CLIENT_BINCMD="${11}"
+SCARAB_HASH="${12}"
 
 for token in $ENVVAR;
 do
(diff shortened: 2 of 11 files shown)


hlitz:
adding @chandrasish961 

why do we need to commit/stash? what happens if we don't?

what do you mean with: Checks out any missing git ?

in general I like most of this PR, but it will be common to develop, test run without committing inbetween. how can we support this use case?

FlyingFish800:
I implemented the changes we discussed in the last meeting.
The git checkout/stashing has been removed. When `scarab_build` is not none, make is run on the current state of the scarab repository. When the resulting binary differs from the cached `scarab_current` binary the cached current binary is updated, and a new scarab binary for that githash is stored in the cache. New entries have a post-fix index that is one greater than the previous maximum index for that githash. The indices count up as follows:
1st: `scarab_<hash>`
2nd:`scarab_<hash>_1`
3rd:  `scarab_<hash>_2`
...
To choose a newer scarab build for a given hash (say `scarab_<hash>_2`) you would add the `_2` postfix to the scarab_githash entry. For example, to choose the third build of the most recent scarab commit, you would have the following in your descriptor:
`"scarab_githash": "6a4e29a_2"`
Additionally, tracing has been updated to use the scarab binary cache. It will always use scarab_current.

5surim:
@FlyingFish800 It looks good to me. Could you resolve the conflicts?
I recently simplified exp.json for the tutorial.

FlyingFish800:
@5surim Great! I resolved the all the conflicts. It's ready to merge now

5surim (approved):
LGTM

PR #206, opened 2025-10-12
Increase SimPoint projection dimension to adapt large workloads like …

moonseoppkim:
Related PR from scarab_ll, https://github.com/litz-lab/scarab_ll/pull/371

Increase SimPoint projection dimension to adapt large workloads like SPEC2017

- Simpoint default dim was 15, but it was not enough to represent the entire workloads according to the error_rate matrix from the PR request msg.

# Overview

## spec2017, 502.gcc_r, full dimensions
image

image

from https://cseweb.ucsd.edu/~calder/papers/JMLR-06-SimPoint.pdf

# Experiments result
image

### DIFF
--- a/common/scripts/run_clustering.sh
+++ b/common/scripts/run_clustering.sh
@@ -19,7 +19,7 @@ else
     maxK=$USERK
 fi
 
-echo "fingerprint size: $lines, maxK: $maxK"
+echo "fingerprint size: $lines, maxk: $maxK"
 # binary search with maxK
 spCmd="$tmpdir/simpoint -maxK $maxK -dim 100 -fixedLength off -numInitSeeds 10 -loadFVFile $FPFILE -saveSimpoints $OUTDIR/simpoints/opt.p -saveSimpointWeights $OUTDIR/simpoints/opt.w -saveVectorWeights $OUTDIR/simpoints/vector.w -saveLabels $OUTDIR/simpoints/opt.l -coveragePct .99 &> $OUTDIR/simpoints/simp.opt.log"
 # search every one with maxK


hlitz (commented):
Can you show a table with the effective number of simpoints for all spec ref apps before and after the patch?
We need to be a bit careful because this will effect simulation time. Essentially we want smallest number of simpoints for each app that provides good enough error rate. We once defined 5% as good enough

moonseoppkim:
> i think large dimension is ok (one time overhead) but we want to reduce maxk so that we have to execute fewer simulations

<meta charset="utf-8"><b style="font-weight:normal;" id="docs-internal-guid-3e6cb845-7fff-a2ac-82f4-c435151a68df">
Trace | 500.perlbench | 502.gcc | 505.mcf | 520.omnetpp | 523.xalancbmk | 525.x264 | 541.leela | 548.exchange2 | 557.xz
-- | -- | -- | -- | -- | -- | -- | -- | -- | --
clusters | 14 | 60 | 12 | 12 | 14 | 14 | 14 | 13 | 12

</b>

This is the current spec2017 rate_int's # clusters.  (/soe/hlitz/lab/traces/spec2017/rate_int)

moonseoppkim:
Hello, @hlitz 

As we discussed on today meeting, I want to add a "-dim 100" option for simpoints3.2 command.

1. The current default dimension value of 15 is outdated because it was originally derived from SPEC2000 workloads. Based on my experiments across different dimensionalities, a dimension of 100 provides a more appropriate trade-off, reflecting the increased complexity of modern benchmarks’ BBV structures.

2. SimPoint 3.2 reduces the original BBV vectors using a random projection matrix, and projecting into a higher dimension does not guarantee better clustering quality. Since the projection matrix is initialized with random values, overly large dimensions can introduce noise rather than additional useful information.
(I attached experiment result matrix as well)

This is the experiments results with several dimensions.

## Error rate
Screenshot 2025-10-29 at 9 10 30 PM

## Execution time
image


## 502.gcc_r deepdive
+ FYI) A very large dimension is not necessarily beneficial.

Screenshot 2025-10-29 at 9 12 51 PM

Screenshot 2025-10-29 at 9 12 19 PM

moonseoppkim:
> @moonseoppkim This looks good to me. Can you maybe run the same methodology for our other workload suites as well? We need to retrace everything anyways I assume.

Sure, I'm planning to integrate the GPU based simpoints script into scarab-infra.
(+ just user can choose a methodology they want between CPU based original or GPU simpoints, based on user's machine.)

Once, I merge the new feature into this repo. I can retrace other workloads too.

+ Ideally, I hope MAV will be considered at the next retracing.

hlitz:
@5surim do you support merging this patch as well?

5surim (approved):
LGTM

moonseoppkim:
> > @5surim do you support merging this patch as well?
> 
> Yes, it looks good to me as well. Do we have a plan to replace the existing SPEC traces as well? @moonseoppkim Have you already collected the new traces for SPEC? Then what is the size of their warm-up? We want to make sure each trace includes at least 50M warmup and 10M main.
> 
> Thanks, Surim

I used a default warmup cnt("10M") for my new spec simpoints.

```python
def cluster_then_trace(workload, suite, simpoint_home, bincmd, client_bincmd, simpoint_mode, drio_args, clustering_userk, manual_trace=False):

    chunk_size = 10000000
    seg_size = 10000000

            # this will limit the amount of warmup that can be done during the simulation
            warmup = seg_size
``` 

I think we have two options for re-tracing the current lab worklods.

1. GPU based full dimension simpoints -> re-tracing
2. GPU based full dimension simpoints + MAV implementaion -> re-tracing

I think the 2nd option sounds like more ideal, 
However, it takes relatively longer than first option in order to implement and verify the results practically.

Thanks.
Best regards,

PR #220, opened 2025-10-22
--interactive goes the base exp path for debugging by default

5surim:
`./sci --interactive <descriptor>` launches an interactive shell and goes directly to the base path of the experiment described in the descriptor to make debugging easier.

Before this PR,
`./sci --interactive <descriptor>` goes to /home/<USER> in the container.
After this PR,
it goes to /home/<USER>/simulations/<experiment> in the container.


FlyingFish800 (approved):
LGTM!

PR #224, opened 2025-10-29
Improve failed simulation tracking

5surim:
- Before exporting stats, we now enumerate every simpoint/config path from the descriptor and abort if any run directory lacks inst.stat.0.csv instead of checking new and old logs (stat_collector.py)
- Add helper to compare descriptor expectations with the CSV, automatically re-run the stat collector when worloads/configs are missing, and reload data after a refresh (sci)
- --status uses a set for error logs, flags segfaults explicitly, and verifies that each log's simpoint directory actually contains inst.stat.0.csv before making it successful (slurm_runner.py)
- rebuild_scarab now always rebuilds with scarab_build or "opt" and copies the new binary into cache only when it differs

### DIFF
--- a/scarab_stats/stat_collector.py
+++ b/scarab_stats/stat_collector.py
@@ -2,7 +2,6 @@
 
 import argparse
 import json
-import os
 import sys
 from pathlib import Path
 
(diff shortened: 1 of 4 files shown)


5surim:
@FlyingFish800 Could you review this PR?

PR #229, opened 2025-11-05
Add docker start step to ./sci --init

5surim:
`./sci --init` had only checked whether Docker is installed and configured the permissions.
It is missing docker start process.
This PR adds the step to start a Docker daemon into `./sci --init`


moonseoppkim (approved):
LGTM

PR #230, opened 2025-11-05
Clean up the failed experiment only csv, out, in, and log files.

5surim:
Sometimes the simulation needs some static information where it is placed in the directory where we should not clean up the whole directory with rm -rf


FlyingFish800 (approved):
LGTM!

PR #236, opened 2025-11-14
Fix slurm maintanance jobs

hlitz:
Slurm jobs cannot be utilzed to perform maintanance tasks such as checking image
availability and node health as this causes livelock and inefficient resource usage.
This PR eliminates all srun/sbatch calls except for the one submitting scarab jobs.
Checking node availability can be performed locally by checking "sinfo -Nel". Docker
image availablity check (and potential pull) is performed as part of every job
submission (we now submit a script that atomically checks image availability and
then submits the job). Submitting jobs is now non-blocking, as such, we cannot run
finish_sim to i.e. perform stats collection. This needs to be performed as part of
the visualize step. Also fixes a bug in status to determine running jobs.
This PR improves resource utiliation and overall simulation speed by 5x.

Limitations:
- The new script assumes that any node that is not in "drain" state has docker daemon
running. The proper way to ensure this is have e.g. a cron job that ensures this is
the case when a node is added to idle state (TODO).
- Not tested with trace mode

### DIFF
--- a/scripts/local_runner.py
+++ b/scripts/local_runner.py
@@ -227,7 +227,7 @@ def run_simulation(user, descriptor_data, workloads_data, infra_dir, descriptor_
                 if scarab_binary not in scarab_binaries:
                     scarab_binaries.append(scarab_binary)
 
-        scarab_githash, image_tag_list = prepare_simulation(user, scarab_path, scarab_build, descriptor_data['root_dir'], experiment_name, architecture, docker_prefix_list, githash, infra_dir, scarab_binaries, interactive_shell=False, available_slurm_nodes=[], dbg_lvl=dbg_lvl)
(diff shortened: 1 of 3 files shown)


FlyingFish800:
In testing, I found that when prepare_docker_image needs to prepare a docker image, it generates extra lines that interfere with the --status command. It was causing these runs to appear as having errors when they did not. I added 'BEGIN prepare_docker_image' and 'END prepare_docker_image' to the beginning and end of the prepare_docker_image output. I added logic to --status to ignore all lines in between the begin and end lines when they are present. There is also 'FAILED prepare_docker_image' when this process fails, which adds that run to the list of failed runs.

Otherwise, simulation and tracing both worked in my testing, and the implementation looks good to me

hlitz:
not expected

PR #260, opened 2025-12-17
Fixed issue where assert failure would not show up in --status

FlyingFish800:
- Changed behaviour when unable to determine state of logfile to assume failure. Previously we would assume unknown log files were running
- We now check slurm queue to assign 'running' status
- Log files now print 'Error' when scarab exits with non-zero status. Previously it would fail silently

### DIFF
--- a/scripts/slurm_runner.py
+++ b/scripts/slurm_runner.py
@@ -465,8 +465,6 @@ def print_status(user, job_name, docker_prefix_list, descriptor_data, workloads_
                 # Didn't build container, use full contents
                 contents_after_docker = contents
 
-            error = 0
-
             # Seventh line indicates completion but lets double check
             if len(contents_after_docker.split("\n")) > 6 and "Completed Simulation" in contents_after_docker:
                 inst_stat_missing = False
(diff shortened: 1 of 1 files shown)


scripts/slurm_runner.py
```diff
@@ -520,12 +520,29 @@ def print_status(user, job_name, docker_prefix_list, descriptor_data, workloads_
...
+                    skipped += 1
+                    if config in running:
+                        running[config] += 1
+                        continue
+                    else:
+                        print("Should not happen. Marked as failure")
+
+            # Assume it failed
+            failed[config] += 1
+
```
5surim: Is it possible to merge the lines from 504~518 into this logic (`Segmentation fault` and `Error`)?
Also, if they are failed jobs, I think they should be added to `error_runs` as it does for `Segmentation fault` and `Error`.

5surim (commented):
Hi @FlyingFish800,

Thanks for the patch. Do you think we can simplify the logic to catch the failed jobs?

Surim

scripts/slurm_runner.py
FlyingFish800: I have simplified the logic, and added the missing `error_runs` line

PR #262, opened 2025-12-28
Cleaned up unused checks, fixed queue check, added nfs syncing

### LINKED ISSUE #261: Improve consistency and clean code for --status
Clean up the implementation of --status
- Remove unnecessary checks (especially relating to stats runner)

Sync NFS before running any file-based checks in --status

FlyingFish800:
- Removed unused code and redundant checks
- Added syncing of logs to NFS after commands which write to log file complete
- Added syncing of local directory to NFS before --status checks log files
- Fixed logic to check squeue for pending/queued simpoints, and incresaed robustness

Closes #261

### DIFF
--- a/scripts/slurm_runner.py
+++ b/scripts/slurm_runner.py
@@ -477,6 +477,8 @@ def print_status(user, job_name, docker_prefix_list, descriptor_data, workloads_
                             if any(list(map(lambda x: x.endswith(".csv"), os.listdir(sim_dir)))):
                                 completed[config] += 1
                                 continue
+                # This error can be triggered by either 1) a NFS syncing error, or 2) SLURM not capturing a failed simulation
+                err("Stat files not generated, despite being completed with no errors.", 1)
 
             # Error detected, or undefined stat (assume it failed)
             error_runs.add(root_logfile_directory+file)


scripts/slurm_runner.py
```diff
@@ -466,38 +433,11 @@ def print_status(user, job_name, docker_prefix_list, descriptor_data, workloads_
                 # Didn't build container, use full contents
                 contents_after_docker = contents
 
-            # Seventh line indicates completion but lets double check
-            if len(contents_after_docker.split("\n")) > 6 and "Completed Simulation" in contents_after_docker:
```
hlitz: are we now only checking the slurm log file? if we perform the sync can we leave this check in to verify that the stats have been written successfully?

FlyingFish800: We are now only checking the slurm log file. We can leave this check if sync the output directory for that simpoint. I will re-implement the check to verify the stats files have been written correctly

hlitz:
@FlyingFish800 can you add the comment that if the coompleted check fails it may be either NFS bug or that slurm did not capture failed simulation? then we can merge

PR #266, opened 2026-01-08
Make the interactive shell always accessible even with scarab build failure

5surim:
scarab build always clean the binary first (make clean) and build again


hlitz (approved):
LGTM

PR #273, opened 2026-01-14
Fix sim PARAMS per scarab hash

5surim:
When running --sim with different scarab hashes (./sci --sim multilev_btb), it checks out the specified git hash and builds scarab to prepare the binary. However, it still uses the same PARAMS.<architecture> file although the different git hashes have differences in `PARAMS.<architecture>`. It should also copy the correct `PARAMS.in` for the corresponding githash not only the binaries.

Fix:
- Export PARAMS.<arch> from each requested scarab git hash into the sim tree
- Update memtrace/pt/exec run scripts to pick hash-specific PARAMS files when present


hlitz (approved):
LGTM

PR #191, opened 2025-08-17
spec2017 build error

moonseoppkim:
=> ERROR [36/40] RUN mount -t iso9660 -o ro,exec,loop /tmp_home/cpu2017-1_0_5.iso /tmp_home/cpu2017_install 0.4s ------ >

[36/40] RUN mount -t iso9660 -o ro,exec,loop /tmp_home/cpu2017-1_0_5.iso /tmp_home/cpu2017_install: 0.286
 mount: /tmp_home/cpu2017_install: mount failed: Operation not permitted.
 

Hello, Surim

I saw a build error while I was building a spec2017 docker image.
Do you think it is a valid change?
* https://stackoverflow.com/questions/26050899/how-to-mount-host-volumes-into-docker-containers-in-dockerfile-during-build
* This post says the "mount" command is not applicable at the docker image build script.

Thanks.

PR #276, opened 2026-01-18
spec2006 workloads setup

moonseoppkim:
spec2006 workloads setup

### DIFF
--- a/workloads/spec2006/Dockerfile
+++ b/workloads/spec2006/Dockerfile
@@ -11,6 +11,16 @@ RUN DEBIAN_FRONTEND=noninteractive apt-get install -y \
     lsb-release \
     software-properties-common
 
+# SPEC CPU2006 Build Configuration
+# 
+# Build command:
+#   runspec --config=Example-linux64-amd64-gcc43+ --action=build --size=ref all
+#
+# Build error fixes:
+#   - 400.perlbench: Add -std=gnu89
+#   - 416.gamess:    Add -std=legacy
+#   - 447.dealII:    Add -std=c++03
+#   - 450.soplex:    Add -std=c++03
 WORKDIR $tmpdir
 RUN mkdir -p $tmpdir/application
 


5surim (changes requested):
Hi @moonseoppkim ,

Thanks for the patch!

A couple of comments and questions:

1. Could you split this PR into two?
- one PR for SPEC2006-specific changes
- another PR for non-SPEC2006 / general changes (Dockerfile.common and run_simpoint_trace.py)
This would make review and future maintenance much easier.

2. Do we actually need to set up SPEC2006 workloads in infra?
If it’s required, could you briefly explain the motivation / use case?

3. About memtrace.cfg:
Where was this file sourced from?
Which lines were manually modified (and why)?

4. Also, could we download the config at build/setup time (e.g., via wget) rather than checking in the full file?
Adding the full upstream file into the repo feels heavy and generally isn’t ideal for version management. If we need reproducibility, pinning a URL + checksum might be a cleaner approach.

Thanks,
Surim

hlitz:
regarding 2. I asked Moon to set it up. We need it for some Samsung project.

moonseoppkim:
Hi @5surim,

This is an early draft. I'll let you know when it's ready.

5surim:
> > > Hi, @5surim
> > > I understand your suggestion, but we have an issue implementing it this way.
> > > We merged #191 last year, which introduced a new config for mounting the application folder (including the pre-built spec folder). We don't build all spec workloads during Docker image build time—we just mount the pre-built spec folder.
> > > So, I don't think your suggestion is feasible given our current setup.
> > > Thanks.
> > 
> > 
> > I think the mounted application folder should contain the .cfg file. If so, it should be feasible to generate the .cfg file at runtime and perform string replacements from workload_root/user_entrypoint.sh.
> > Surim
> 
> Yes, that might be a possible option.
> 
> However, it takes at least 20-30 minutes every time we run containers to rebuild all workloads from scratch. I think the main purpose of #191 was to eliminate that loading time by simply mounting the target folder.

Why are the two commands (cp and sed) introducing 20 mins building time?
I meant we can use the current mount mechanism (#191) but with an additional step in entrypoint.sh.
Also, you can even skip the two commands if the mounted folder already includes the memtrace.cfg file already created by the very first container.

5surim (approved):
LGTM

Issue #264, opened 2026-01-05
[BUG] extracting simpoints with cluster_then_trace


moonseoppkim:
Hello, @hlitz and @5surim 

I think the current "cluster_then_trace" implementation is too inaccurate to extract representative simpoints.
This is my observations,

## Issue Overview
1. BBV are constructed based on fetched instruction for both ways(trace_then_cluster & cluster_then_trace). [REP as a single instruction],
2. "cluster_then_trace" set the roi_begin inaccurately with trace_after_instr parameter. [REP as multiple instructions for "trace_after_instrs"]
FYI) We are currently using "cluster_then_trace" type simpoints only for spec2017.

## 1. "Basic Blocks" treats REP instruction as a single instruction regardless trace_then_cluster or cluster_then_trace.
```c++
1. trace_then_cluster

"--use_fetched_count=1 ", we use this option for trace_then_cluster for building a BBV.
REP counts "1".

* bbv building command with scarab
  scarabCmd="$SIMPOINTHOME/scarab/src/scarab --frontend memtrace \
            --cbp_trace_r0=$TRACEFILE \
            --mode=trace_bbv_distributed \
            --segment_instr_count=$SEGSIZE \
            --memtrace_roi_begin=$(( $segmentID * $SEGSIZE + 1 )) \
            --memtrace_roi_end=$(( $segmentID * $SEGSIZE + $SEGSIZE )) \
            --trace_bbv_output=$OUTDIR/fingerprint/pieces/segment.$segmentID \
            --trace_footprint_output=$OUTDIR/fingerprint/footprint_pieces/segment.$segmentID \
            --use_fetched_count=1 \
            &> sim.log"
``` 

```c++
2. cluster_then_trace

scarab-infra/fingerprint_src/fpg.cpp
* BBV building source for cluster_then_trace.

dr_client_main(client_id_t id, int argc, const char *argv[])
{
...
    // dr_register_bb_event(event_basic_block);
    if (!drmgr_register_bb_app2app_event(event_bb_app2app, NULL))
        DR_ASSERT(false);
    if (!drmgr_register_bb_instrumentation_event(event_app_analysis, /*event_app_instruction*/ NULL, NULL))
        DR_ASSERT(false);
...
}


event_bb_app2app()
{
...
    https://dynamorio.org/group__drutil.html#gae38961d42fc285d7f9034c4b02690cae
[...]

PR #252, opened 2025-12-10
Make USE_FETCHED_COUNT=1 run fetched instruction only simulation

moonseoppkim:
I think both trace_then_cluster and cluster_then_trace simulation expect fetched-instruction oriented result.

image

exit_after_tracing -> after tracing the specified number of references, the process is exited with an exit code of 0. The reference count is approximate

number of ifetch + number of write = 10M (exit_after_tracing limit)
number of instruction = number of ifetch

* dynamorio, parameter documentation
https://dynamorio.org/sec_drcachesim_ops.html

use_fetched_count feature history : https://github.com/litz-lab/scarab-infra/pull/133

PR #279, opened 2026-01-19
Fix bug in cluster_then_trace for setup roi

moonseoppkim:
## Changes
```
1. Use a customized DynamoRIO for Scarab
   1) trace_after_instrs: count only fetched instructions
      - The fpg.cpp uses the "count_app_instrs" internal function to count fetched instructions for each Basic Block generated by DynamoRIO
      - However, we skip some basic blocks that have the same PC as the previous block in order to construct BBV
      - Therefore, we need to add "prev_pc" checking logic to DynamoRIO's trace_after_instrs & trace_for_instrs to count instructions the same way as fpg.cpp
   2) trace_for_instrs: count only fetched instructions
2. max_bb_instrs count (4096 -> 4095)
  1) ASSERT FAILURE: /dynamorio_full_original/dynamorio/clients/drcachesim/tracer/tracer.cpp:2608: max_bb_instrs < uint64(1) << 12 ()
  2) max_bb_instrs should be lesser than 2^12(4096), and we need to use same parameters for both executions, so that DynamoRIO can generate same basic blocks.
  3) default value for "max_bb_instrs" is 1024.
``` 

We need to merge #252 together with newly collected traces.

Note : https://github.com/litz-lab/scarab-infra/issues/264

### DIFF
--- a/common/Dockerfile.common
+++ b/common/Dockerfile.common
@@ -51,9 +51,23 @@ RUN mkdir -p $tmpdir && chmod -R 755 $tmpdir
 WORKDIR $tmpdir
 
 # DynamoRIO package
-RUN cd $tmpdir && wget https://github.com/DynamoRIO/dynamorio/releases/download/release_10.0.0/DynamoRIO-Linux-10.0.0.tar.gz && tar -xzvf DynamoRIO-Linux-10.0.0.tar.gz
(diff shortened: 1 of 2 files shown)


hlitz (approved):
LGTM

PR #284, opened 2026-01-22
Unify conda path resolution for manual runs

5surim:
- prefer CONDA_EXE and ~/miniconda3/bin/conda before PATH
- resolve scarabinfra env via conda env list
- align serve_jupyter.sh conda lookup with sci behavior

This should fix the following error in the manual mode.

EnvironmentLocationNotFound: Not a conda environment: /soe/surim/.conda/envs/scarabinfra


hlitz (approved):
LGTM

PR #299, opened 2026-01-27
Fixed slurm job queue log to prevent spamming when many configs are used

FlyingFish800:
Puts all `Submitting jobs: <...>` messages on one line to prevent spamming the terminal


FlyingFish800:
The newline before "submitted all jobs" is there to separate it from the line of "Submitting jobs: <...>" statements, which are always followed by a single space. Without that newline, it would print on the same line as those statements. 
Here is a demonstration of what it looks like with this newline:
```
Submitting jobs...
Submitting jobs: 2 Submitting jobs: 4 Submitting jobs: 6 Submitting jobs: 8 Submitting jobs: 10 
Submitted all jobs
```
If it would be better to print the "Submitted all jobs" message on the same line, I can remove the newline character

5surim (approved):
LGTM

PR #300, opened 2026-01-27
Slurm options

FlyingFish800:
Added `slurm_options` field for each config (traces and simulations). This is a string used for passing options directly to squeue. For example, `"--mem 10G"` can be used to allocate 10GB of ram, rather than the default 4GB. It can also be used to select which node (or partition) a job should be run on, which will be useful for GPU based tracing.

### DIFF
--- a/scripts/utilities.py
+++ b/scripts/utilities.py
@@ -1381,9 +1381,7 @@ def print_simulation_status_summary(
                 error_runs.add(log_path)
                 continue
 
-<<<<<<< HEAD
             prep_err = 0
-=======
             workload_parts = workload_path.split("/")
             sim_dir = Path(descriptor_data["root_dir"]) / "simulations" / descriptor_data["experiment"] / config
             sim_dir = sim_dir.joinpath(*workload_parts, cluster_id)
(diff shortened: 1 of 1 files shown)


FlyingFish800:
Hello @5surim, sorry I needed to update the --status command to catch the OOM killed sps, and print the configs which had these errors. I made some small changes to the status command in the latest commit. Can you please re-review this PR? Thanks!

json/exp.json
5surim: Could you update the description?
Please describe how to check the available Slurm options and list the current ones.

json/top_simpoint.json
```diff
@@ -32,7 +32,8 @@
   "configurations": {
       "this_pr": {
         "params": "--bp_mech tage64k",
-        "binary": "scarab_current"
+        "binary": "scarab_current",
+        "slurm_options": "--mem 4G"
```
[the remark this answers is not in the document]
FlyingFish800: Thanks for the comments! It does correctly ignore the slurm options if the workload manager is manual. I added a note that an empty string works as a default value, and updated top_simpoint_dbg.json

json/exp.json
5surim: Can you move it to _configurations so that we don't need to have this long description for every slurm_options? Also, we always want to keep exp.json up-to-date, so please add the description to exp.json as well.

PR #301, opened 2026-01-27
Rewrite scarab_current as a symlink to a hash-specific binary

5surim:
If scarab_current is a non-symlink (old infra) and the rebuilt binary is identical, it still rewrites scarab_current as a symlink.


5surim:
> I tested with a non-symlinked scarab_current binary, and the patch worked as intended. LGTM!

Thanks @FlyingFish800 for testing it!

PR #314, opened 2026-02-21
Improve --kill, --clean, --status

5surim:
## Problems

- `./sci --sim` could start while stale containers with deterministic names still existed, causing Docker name conflicts.
- Wrapper scripts could leave orphan containers on abnormal exits (cancel/signal/interruption), leading to repeated relaunch failures.
- In Slurm mode, `./sci --kill` canceled jobs but did not automatically perform container cleanup.
- `./sci --status` could misreport failures due to:
  - older wrapper logs from previous attempts
  - canceled wrapper logs even when simulation outputs (.csv) were already generated.

### 1) Preflight conflict guard before simulation launch

- Added container-conflict preflight in scripts/run_simulation.py.
- Checks docker ps -a for matching experiment containers (local or Slurm nodes).
- Refuses `--sim` launch when conflicts exist and prints actionable recovery guidance.

### 2) Hardened wrapper cleanup with traps

- Updated generated simulation and trace wrapper scripts in scripts/utilities.py:
  - introduced CONTAINER_NAME
  - added cleanup_container() helper
  - added trap cleanup_container EXIT INT TERM HUP
- Standardized docker commands to use CONTAINER_NAME and explicit cleanup call at script end.

### 3) Slurm --kill now includes cleanup

- In Slurm mode, --kill now performs:
  - scancel of matching jobs
  - container cleanup
  - temporary run-script cleanup
- Implemented via scripts/run_simulation.py and scripts/slurm_runner.py (kill_jobs() returns cleanup intent).

### 4) Status classification fixes

- --status in scripts/utilities.py now:
  - tracks the latest wrapper log per scenario (config + workload_path + cluster_id)
  - counts cancelled wrapper runs as Completed when corresponding .csv output exists
  - keeps non-cancel errors classified as failed (unchanged)
- Removed misleading old-run warning and updated summary wording to indicate latest-log-based reporting.

## Result

- Relaunch failures from stale container collisions now happen rarely. They should only occur in hard-failure scenarios:
  - node crash/reboot
  - hard kill (SIGKILL)
  - node unreachable during cleanup
  - (less common) Docker daemon hang/restart during cleanup
  - (less common) permission/transport failures on remote cleanup commands
- Even when this happens, the preflight guard blocks relaunch and directs users to clean stale containers (for example, ./sci --clean <descriptor>), improving cleanup coverage for interrupted runs.
- Slurm kill flow is now operationally complete in normal cases, without requiring immediate manual clean.
- Status output better reflects real outcomes for cancel-but-finished scenarios while still surfacing real errors.

### DIFF
--- a/scripts/utilities.py
+++ b/scripts/utilities.py
@@ -94,7 +94,16 @@ def write_json_descriptor(filename, descriptor_data, dbg_lvl = 1):
 
 def run_on_node(cmd, node=None, **kwargs):
     if node != None:
-        cmd = ["srun", f"--nodelist={node}"] + cmd
+        # Use fast-fail scheduling for management commands so they do not
+        # block behind busy-node resource allocation.
+        cmd = [
+            "srun",
+            f"--nodelist={node}",
+            "--nodes=1",
+            "--ntasks=1",
+            "--cpus-per-task=1",
+            "--immediate=5",
+        ] + cmd
     return subprocess.run(cmd, **kwargs)
 
 def validate_simulation(workloads_data, simulations, dbg_lvl = 2):


hlitz (commented):
Thanks Surim this looks great.

@FlyingFish800 can you test this patch if you don't see any issues please approve.

PR #319, opened 2026-03-03
Fixed issue where slurm options would prevent sbatch from running

FlyingFish800:
Without this code, there is no space between `-c 1` and other arguments. This logic is required because extra spaces cause sbatch to fail when run through subprocess.run()


5surim (approved):
LGTM. I also experienced this issue. Thanks for fixing it.

PR #321, opened 2026-03-04
Fix stats library

5surim:
When we introduce new stats in scarab, --visualize ignores them by using the first configuration to get the stats information. This limits not only collected_stats.csv (missing new stats) but also visualization of new stats.

With this PR,
- do not use the first configuration for stats, but union of them so we don't miss new stats in new config/scarab
- add rotation to the stacked bars so that x-axis labels are not overlapped.

### DIFF
--- a/scripts/utilities.py
+++ b/scripts/utilities.py
@@ -8,6 +8,7 @@ import subprocess
 import re
 import shlex
 import shutil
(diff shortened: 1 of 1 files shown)


hlitz (approved):
LGTM

PR #325, opened 2026-03-11
Ignore untracked files in the Scarab dirty-check logic

5surim:
Previously, we rebuilt every time if there was an untracked file in scarab repo.
Ignore the untracked file during the dirty check.


hlitz (approved):
LGTM

PR #326, opened 2026-03-13
Mem collect

5surim:
### Problem

Slurm simulations used a single manually-configured `--mem` value per configuration in `slurm_options`. This required users to guess a safe memory limit upfront: too low caused OOM kills that wasted cluster time, too high wasted cluster resources. There was also no visibility into per-workload memory consumption across configurations, and no enforcement of consistent JSON formatting across tracked descriptor files.

## Auto memory scheduling (`scripts/slurm_runner.py`)

Per-simpoint `--mem` is now computed automatically on every submission. The formula:
```
mem_mb = int(base_memory_mb * 1.2) + memory_overhead_mb
```
- `base_memory_mb`: measured RSS of the baseline config for that specific simpoint, stored in `workloads_db.json` by `--collect-mem`
- 1.1×: 10% headroom over the measured baseline
- `memory_overhead_mb`: fixed per-config overhead declared in the descriptor for configs that allocate extra Scarab structures (e.g. larger BP tables) on top of the baseline

When no measurement exists yet (first run), it falls back to `mem.fallback_mb + memory_overhead_mb`. `--mem` in `slurm_options` is now deprecated (If present, it is stripped with a warning.).

##  Memory collection (`sci --collect-mem <descriptor>`)

New CLI command to harvest RSS measurements from completed Slurm jobs:

1. Scans `logs/job_*.out` using `iter_latest_job_logs` — deduplicates by highest job ID so re-runs and OOM-killed jobs are superseded by the latest successful run
2. Queries `sacct --format=JobID,MaxRSS` (without -X) and takes the max across all job steps including `.batch`, where MaxRSS is actually populated
3. Attributes each measurement to its simpoint via the `Running {config} {suite}/{subsuite}/{workload} {cluster_id}` first-line header written by the job script
4. Writes `base_memory_mb` (the baseline config's RSS) back into `workloads_db.json` per simpoint, and records `_scarab_sim_githash` at the subsuite level to track which Scarab binary the measurement was taken with
5. Writes `memory_stats.json` into the experiment directory with per-config RSS for all measured simpoints

The baseline config used for `base_memory_mb` is controlled by `mem.collection_baseline` in the descriptor (defaults to `visualize.baseline`).

## Memory visualization (`sci --visualize`)

  When `memory_stats.json` exists, --visualize appends a markdown table showing RSS averaged across simpoints per workload, with percentage deltas vs. the baseline config. Configurations exceeding 20% RSS increase are flagged in red.

## Descriptor schema additions

  Three new fields, following the existing `visualize` / `perf_analyze` pattern:

| Field                   | Location             | Purpose                                                                                                      |
|-------------------------|----------------------|--------------------------------------------------------------------------------------------------------------|
| memory_overhead_mb      | per core entry       | Fixed MB overhead for extra Scarab structures allocated by this config on top of the simpoint's baseline RSS |
| mem.collection_baseline | descriptor top-level | Which config's RSS --collect-mem stores as base_memory_mb; defaults to visualize.baseline                    |
| mem.fallback_mb         | descriptor top-level | --mem limit used for simpoints with no measurement yet; set conservatively high to avoid OOM on first run    |

  `exp.json`, `top_simpoint.json`, and `top_simpoint_dbg.json` are updated accordingly. `--mem` is removed from all slurm_options entries.

## Shared log-parsing API (`scripts/utilities.py`)

  Three new module-level utilities extracted from duplicated logic that existed independently in `--status` and `--collect-mem`:

  - `parse_job_log_header(first_line)` — parses "Running {config} {suite}/{subsuite}/{workload} {cluster_id}", returns a 5-tuple or None
  - `get_experiment_logs_dir(descriptor_data)` — returns `Path(root_dir)/simulations/{experiment}/logs`
  - `iter_latest_job_logs(logs_dir)` — scans `job_*.out`, deduplicates by highest job ID, yields `(job_id_int, log_path, config, suite, subsuite, workload, cluster_id_str)`

  `print_simulation_status_summary` is refactored to a single pass over `job_*.out` using these helpers. A notable fix: the literal `job_%j.out` placeholder file (touched by the runner before each `sbatch` call) previously fell into the mtime-ded
up path and was counted in the log total. The new code counts it
  explicitly as an unparseable entry, preserving the `log_count_offset=1` assertion in `slurm_runner.py` that was calibrated against it.

 ## CI enforcement (`.github/workflows/check-json-validity.yml`)

  New "Validate JSON formatting" step that enforces `indent=2, separators=(',', ':')` (no spaces after : or ,) for `exp.json`, `top_simpoint.json`, `top_simpoint_dbg.json`, `workloads_db.json`, and `workloads_top_simp.json`. Fails with a fix command if any file is non-compliant.

  ---
  `workloads/workloads_db.json`

  - "_scarab_sim_githash": null added at every subsuite level
  - "base_memory_mb": null added to every simpoint entry
  - Entire file reformatted to canonical indent=2, separators=(',', ':') JSON

  ---
## Typical workflow after this PR
```
  # First run: uses mem.fallback_mb (set conservatively) for all simpoints
  ./sci --sim my_experiment

  # After jobs complete: harvest actual MaxRSS from Slurm accounting
  ./sci --collect-mem my_experiment

  # Subsequent runs: use measured base_memory_mb * 1.1 + overhead per simpoint
  ./sci --sim my_experiment

  # Visualize IPC plots + memory delta table across configs
  ./sci --visualize my_experiment
```

### DIFF
--- a/scripts/run_simulation.py
+++ b/scripts/run_simulation.py
@@ -100,6 +100,19 @@ def verify_descriptor(descriptor_data, workloads_data, open_shell = False, dbg_l
 def open_interactive_shell(user, descriptor_data, workloads_data, infra_dir, dbg_lvl = 1):
     experiment_name = descriptor_data["experiment"]
     scarab_path = descriptor_data["scarab_path"]
+    scarab_binaries: List[str] = []
+    configurations = descriptor_data.get("configurations") or {}
+    if isinstance(configurations, dict):
+        for config in configurations.values():
+            if not isinstance(config, dict):
+                continue
+            binary = config.get("binary")
+            if binary:
+                binary_name = str(binary)
+                if binary_name not in scarab_binaries:
+                    scarab_binaries.append(binary_name)
+    if not scarab_binaries:
+        scarab_binaries = ["scarab_current"]
     try:
         # Get user for commands
         user = subprocess.check_output("whoami").decode('utf-8')[:-1]
@@ -161,7 +174,7 @@ def open_interactive_shell(user, descriptor_data, workloads_data, infra_dir, dbg
                                                 docker_prefix_list,
                                                 githash,
                                                 infra_dir,
-                                                ["scarab_current"],
+                                                scarab_binaries,
                                                 interactive_shell=True,
                                                 dbg_lvl=dbg_lvl)
         except Exception as exc:
--- a/scripts/utilities.py
+++ b/scripts/utilities.py
@@ -414,6 +414,100 @@ def _current_git_ref(scarab_path: str) -> str:
     return branch_name if branch_name else commit_hash
 
 
+def _stash_ref_for_message(scarab_path: str, message: str) -> Optional[str]:
+    try:
+        stash_list = subprocess.check_output(
+            ["git", "stash", "list", "--format=%gd\t%gs"],
+            cwd=scarab_path,
+            text=True,
+        )
+    except subprocess.CalledProcessError:
+        return None
+
+    for line in stash_list.splitlines():
+        ref, _, subject = line.partition("\t")
+        if subject == message:
+            return ref.strip() or None
+    return None
+
+
+@contextmanager
+def _temporary_scarab_checkout(scarab_path: str, target_ref: str, dbg_lvl: int):
+    original_ref = _current_git_ref(scarab_path)
+    stash_ref = None
+    stash_message = f"sci-temp-scarab-build-{os.getpid()}-{int(time.time())}"
+
+    if not _scarab_repo_clean(scarab_path):
+        warn(
+            "Scarab repo has uncommitted changes; stashing them temporarily to build a hash-pinned binary.",
+            dbg_lvl,
+        )
+        try:
+            subprocess.run(
+                ["git", "stash", "push", "--include-untracked", "-m", stash_message],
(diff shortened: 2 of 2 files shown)


5surim:
> @5surim got it, I'll create that pr with the populated database. I ran into another issue with `--status` though, I got the same error as before, this time on line 1193 of `utilities.py`
> 
> ```
>     predef_sim_mode = workloads_data[suite][subsuite][workload]["simulation"]["prioritized_mode"]
>                       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^
> TypeError: string indices must be integers, not 'str'
> ```
> 
> And when using `--visualize`, on line 2415 in `scarab_stats.py`
> 
> ```
>     sim_mode = workloads_data[suite][subsuite][workload]["simulation"]["prioritized_mode"]
>                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^
> TypeError: string indices must be integers, not 'str'
> ```

Ah, good catch. Will add another fix commit here.

FlyingFish800 (approved):
Tested it, and everything works now! LGTM

PR #327, opened 2026-03-13
Accelerate scarab build via sci

hlitz:


FlyingFish800:
Does this work? I tried bumping the make threads like you requested, but I don't think it used more cores. Bohr1 has a default cgroup slice that only allows users to use 8 cores, I believe it was set by Yuanpeng (it's mentioned in #300). I've been trying to increase the cgroup core allotment to 20 but I haven't been successful yet

FlyingFish800:
I believe without the `-j` option it only uses 1 core so this is still an improvement, even if it only uses 8

FlyingFish800:
It seems to work in my testing now, I see activity on all cores when I run build scarab.

@hlitz for future reference, where did you find `cgroup.cfg` on bohr1? I looked for it in `/etc/` and `/sys/fs/cgroup/` but couldn't find a default/general config file

FlyingFish800:
I have updated the cgroups so each user should now have access to 20 cores worth of compute when running make. I now see >20% utilization across all cores when running make with this PR. 
The solution was to update the `/usr/local/bin/limit-new-login.sh` to set each user's cgroup cpu allotment to 20 cores when they login.
Should we merge this PR now?

PR #328, opened 2026-03-14
Mem defaults

FlyingFish800:
Added default memory values to the workload database for all datacenter workloads. Default values were collected using empty parameter values, and the latest scarab
The following simpoints were not collected:
dcperf/django_bench/django_bench_server 21 - Missing instruction (see scarab #464)
dcperf/wdl_bench/fibers_benchmark 2,4,54,76,92,141,151,158,198,229 - Missing traces
dcperf/wdl_bench/random_benchmark 10,65,87,190,220,243,289,325,332 - Missing traces

### DIFF
--- a/workloads/workloads_db.json
+++ b/workloads/workloads_db.json
@@ -1,7 +1,7 @@
 {
   "google":{
     "google":{
-      "_scarab_sim_githash":null,
+      "_scarab_sim_githash":"98dc727e",
       "arizona":{
         "simulation":{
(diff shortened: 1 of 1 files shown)


5surim (approved):
LGTM

hlitz:
It is a critical issue. Right now the cluster is almost unusuable. I really wonder where this huge memory increase comes from and we need to address it. Do we know for sure that the warmup is the culprit here?

PR #339, opened 2026-03-20
Handle docker error runs correctly in `--status`

estodi:
The run that failed due to a docker error was mistakenly handled as a failure of scarab. As a result, --status was trying to display sim.log, which doesn’t even exist when docker failed before running scarab. This PR fixes this rare issue and correctly attributes the cause.

### DIFF
--- a/scripts/utilities.py
+++ b/scripts/utilities.py
@@ -1561,6 +1561,11 @@ def print_simulation_status_summary(
                 error_runs.add(str(log_path))
                 continue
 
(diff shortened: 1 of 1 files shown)


5surim (approved):
LGTM! thanks
I guess you are experiencing the failures from the jobs scheduled on bohr5?

estodi:
> I guess you are experiencing the failures from the jobs scheduled on bohr5?

yes that was the issue

PR #340, opened 2026-03-20
Gray out 0 values in the status table for better visibility

estodi:
This PR makes the `--status` table easier to interpret, especially when it contains many rows.


estodi:
When a table has so many rows, it becomes hard and inefficient to find one-digit non‑zero values among the sea of 0s; the digit 8 in particular blends into the surrounding zeros.

Left: before, Right: after
scarab-infra_pr340

PR #343, opened 2026-03-23
Stop forcing Scarab rebuilds when launching interactive simulation or trace descriptors

5surim:
This change makes ./sci --interactive <descriptor> non-destructive for both simulation and trace descriptors. Interactive mode no longer triggers rebuild_scarab() just because the current Scarab repo is dirty or the cache is out of sync. Instead, it tries to reuse an existing cached scarab_current binary for the selected mode, falls back to the existing repo binary under src/build/<mode>/scarab if present, and otherwise warns and continues without rebuilding.

  The warning behavior is also improved. Interactive launches now check both opt and dbg current-binary status and print why a binary is not considered up to date, along with an explicit rebuild command and the required scarab_build value to place in the descriptor. For example, warnings now tell the user to run ./sci --build-scarab <descriptor> and set scarab_build to opt or dbg as needed. In addition, the simulation interactive path no longer validates the same descriptor twice, which removes the duplicate “valid simulation option” lines.

  Behavior By Option And Build Mode
  scarab_build meanings:

  - null: normalized to opt
  - opt: use src/build/opt/scarab
  - dbg: use src/build/dbg/scarab

  Primary commands:

  - `./sci --build-scarab <descriptor>`
  - `./sci --interactive <descriptor>`
  - `./sci --sim <descriptor>`
  - `./sci --trace <descriptor>`

  Behavior matrix:

  - `--build-scarab`
      - null: builds opt
      - opt: builds opt
      - dbg: builds dbg
      - intent: explicit build path, rebuild allowed
  - `--sim`
      - null: runs with opt
      - opt: runs with opt
      - dbg: runs with dbg
      - rebuild policy: unchanged; scarab_current may be rebuilt if cache/hash/dirty checks require it
  - `--trace`
      - null: runs with opt
      - opt: runs with opt
      - dbg: runs with dbg
      - rebuild policy: unchanged for non-interactive trace runs; rebuild may still happen
  - `--interactive` for simulation descriptors
      - rebuild policy: **no forced rebuild**; reuse cache or repo binary if present, otherwise warn
  - `--interactive` for trace descriptors
      - rebuild policy: now matches simulation interactive; **no forced rebuild**, warn instead

  Important distinction:

  - scarab_build selects the mode directory, not whether rebuilding is enabled.
  - rebuild/no-rebuild is determined by the command path:
      - `--build-scarab:` build intentionally
      - `--sim / non-interactive --trace`: rebuild still allowed
      - interactive simulation/trace: skip rebuild, warn if current binaries are missing or stale

### DIFF
--- a/workloads/workloads_db.json
+++ b/workloads/workloads_db.json
@@ -1,7 +1,7 @@
 {
   "google":{
     "google":{
-      "_scarab_sim_githash":null,
+      "_scarab_sim_githash":"98dc727e",
       "arizona":{
         "simulation":{
           "prioritized_mode":"memtrace",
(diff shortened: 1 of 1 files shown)


hlitz:
what do you mean with: to reuse an existing cached scarab_current ?
does --interactive even need a scarab_current? I think one may want to start an interactive session without having a scarab binary (e.g. code may not compile or one wants to do sth completely different)

PR #360, opened 2026-03-28
For entries in visualize options without _value, append it

### LINKED ISSUE #354: eliminate _value for visualize
visualizing PMUs in JSON requires the addition of _value, e.g. SIM_HOST_WALL_SECONDS_value. Automatically append _value so that user only needs to define SIM_HOST_WALL_SECONDS in json. Attention sometimes e.g. with IPC _value does not seem to be required, maybe try both, ie adding _value or not and try finding the stat.

FlyingFish800:
For any entry in the stat visualization section of the json file, check if the counter name is missing '_value'. If it is, append it.

Also moved 'resolve_stat_name' to a separate function as it was used multiple times.

Closes #354

### DIFF
--- a/sci
+++ b/sci
@@ -2808,6 +2808,24 @@ def run_collect_mem(descriptor_name: str) -> int:
     print(f"Per-config memory data written to {memory_stats_path}.")
     return 0
 
+def resolve_stat_name(stat_name: str, available_stats: List[str]) -> Tuple[str, bool]:
+    """
+    Resolve a requested stat to the name present in the aggregated stats.
+    Returns a tuple of (resolved_name, used_count_fallback).
+    """
+    normalized = str(stat_name)
+    if not available_stats:
+        return normalized, False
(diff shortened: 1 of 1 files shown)


FlyingFish800:
@hlitz I found some stats that do end in _value in scarab's 'power' module. They are a separate type (`FLOAT`) than the `COUNT` variables. Changing their suffixes to _count would not have made sense as the stats are named `ENERGY`, `ENERGY_CORE`, and `TIME`. 

I updated scarab infra to accept these stats without the _value suffix using the same mechanism we use for _count. I also moved this function to be globally defined in `sci`, as it was defined separately twice previously. The latest commit doesn't contain any redundant code from the previous commits.

FlyingFish800:
@hlitz I verified this worked with the `SIM_HOST_WALL_SECONDS_value` stat. I also confirmed there are no other suffixes we might need to cover with this patch. 

This pr is ready for review now

hlitz (approved):
lgtm

PR #375, opened 2026-04-07
[BUG] ASLR settings fix

moonseoppkim:
Hi, @5surim 

As we discussed before, I'm working on the ASLR feature fix.

First, I disabled all ASLR recovery logic since it is a system-wide feature. 
This ensures that the setting is not restored while other simulation jobs (especially tracing) are running.

Additionally, I have a few question about next steps for this PR.

```
1. outside user who supports slurm feature
  1) check available slurm nodes (sinfo -N -h -o "%N")
  2) run "echo 0 > /proc/sys/kernel/randomize_va_space"
    1] however, we need root permissions for individual machines in order to run that command.
    2] if someone don't know their machine's sudo pwd? Is it okay to skip this stage?
3. outside user who doesn't support slurm, run only at local.
  1) check available slurm nodes -> return null
  2) run "echo 0 > /proc/sys/kernel/randomize_va_space", only for local
```

I think disabling ASLR should be persistent across all available machines, so it should be handled once during "./sci --init"

Thanks.
Best regards,

### DIFF
--- a/sci
+++ b/sci
@@ -2010,6 +2010,7 @@ def ensure_aslr_disabled(_: argparse.Namespace) -> Tuple[bool, str]:
                 
                 if node_aslr != '0':
                     print(f"\n[!] ERROR: ASLR is enabled on Slurm node [{node}] (value: {node_aslr}).")
+                    print(f"    Please run: srun -w {node} sudo sh -c 'echo 0 > {aslr_path}'")
                     sys.exit(1)
             except Exception:
                 continue
--- a/sci
+++ b/sci
@@ -1985,6 +1985,45 @@ def maybe_check_claude_cli_auth(_: argparse.Namespace) -> Tuple[bool, str]:
     )
 
 
+def ensure_aslr_disabled(_: argparse.Namespace) -> Tuple[bool, str]:
+    aslr_path = "/proc/sys/kernel/randomize_va_space"
+    aslr_file = Path(aslr_path)
+
+    all_nodes = []
+    try:
+        node_output = subprocess.check_output(['sinfo', '-h', '-o', '%N'], 
+                                              universal_newlines=True, stderr=subprocess.DEVNULL).strip()
+        if node_output:
+            all_nodes = subprocess.check_output(['scontrol', 'show', 'hostnames', node_output], 
+                                                universal_newlines=True).splitlines()
+    except (FileNotFoundError, subprocess.CalledProcessError):
+        pass
+
+    if all_nodes:
+        print(f"Detected Slurm cluster. Checking ASLR on {len(all_nodes)} nodes...")
+        for node in all_nodes:
+            try:
+                node_aslr = subprocess.check_output(
+                    ['srun', '-w', node, '-N1', '-n1', 'cat', aslr_path],
+                    universal_newlines=True, stderr=subprocess.DEVNULL, timeout=3
+                ).strip()
+                
+                if node_aslr != '0':
+                    print(f"\n[!] ERROR: ASLR is enabled on Slurm node [{node}] (value: {node_aslr}).")
+                    sys.exit(1)
+            except Exception:
+                continue
+        return True, f"ASLR is disabled across all {len(all_nodes)} Slurm nodes."
+    
+    else:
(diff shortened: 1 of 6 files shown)


5surim (dismissed):
@moonseoppkim 

yes, please work on sci --init.
BTW, should we add an assertion for randomize_va_space?

Thanks,
Surim

moonseoppkim:
Hi, @5surim 

```
scripts/run_trace.py:201:               
 os.system(f"docker exec --privileged {docker_container_name} /bin/bash -c \"echo 0 | sudo tee /proc/sys/kernel/randomize_va_space\"")

scripts/utilities.py:1461:                
f.write("docker exec --privileged $CONTAINER_NAME /bin/bash -c \"echo 0 | sudo tee /proc/sys/kernel/randomize_va_space\"\n")

scripts/utilities.py:1541:            
f.write("docker exec --privileged $CONTAINER_NAME /bin/bash -c \"echo 0 | sudo tee /proc/sys/kernel/randomize_va_space\"\n")
```

The current implementation set the "ASLR" to "0" by utilizing "--pervileged" container.
So, we don't need to worry about "randomize_va_space" at "sci --init" phase.

Thanks.

moonseoppkim:
Hi, @hlitz 

I suspect that the traces previously collected via scarab-infra have an minor?? issue related to ASLR.

When I collected the spec2017_v2, I ran all tracing only from ohm server simultaneously,

The ASLR settings were inconsistently toggled during the parallel execution of tracing jobs, 
which likely caused unintended interference between different workloads.

+FYI)
###  ASLR disabled("0"), when they started
```
scripts/run_trace.py:201:               
 os.system(f"docker exec --privileged {docker_container_name} /bin/bash -c \"echo 0 | sudo tee /proc/sys/kernel/randomize_va_space\"")

scripts/utilities.py:1461:                
f.write("docker exec --privileged $CONTAINER_NAME /bin/bash -c \"echo 0 | sudo tee /proc/sys/kernel/randomize_va_space\"\n")

scripts/utilities.py:1541:            
f.write("docker exec --privileged $CONTAINER_NAME /bin/bash -c \"echo 0 | sudo tee /proc/sys/kernel/randomize_va_space\"\n")
```

### ASLR back("2, default"), when jobs done
```
scripts/run_perf.py:57:                
os.system("echo 2 | sudo tee /proc/sys/kernel/randomize_va_space")

scripts/run_trace.py:196:                
os.system("echo 2 | sudo tee /proc/sys/kernel/randomize_va_space")

scripts/run_trace.py:203:                    
os.system("echo 2 | sudo tee /proc/sys/kernel/randomize_va_space")

scripts/slurm_runner.py:758:        
os.system("echo 2 | sudo tee /proc/sys/kernel/randomize_va_space")

scripts/local_runner.py:478:        
os.system("echo 2 | sudo tee /proc/sys/kernel/randomize_va_space")

scripts/utilities.py:1769:       
 os.system("echo 2 | sudo tee /proc/sys/kernel/randomize_va_space")
```

Therefore, the current traces are not ideally for making correct comparisons between memtrace vs. pin.
Do you think we need to re-collect traces again after finalizing this PR?

Thanks.

moonseoppkim:
@5surim, Could you re-approve please?

5surim:
Yes, I know we set it when we run sim/trace, but I think this should not be silent.
So any user should be aware that this is set when running `./sci --init`.
So removing the revert step is ok, but setting silently is not ok.

moonseoppkim:
Hi, @5surim 

I add an additional assertion for both cases (slurm or manual).

### PASS
```
==> Check ASLR setting
Detected Slurm cluster. Checking ASLR on 6 nodes...
   ASLR is disabled across all 6 Slurm nodes.
```

### FAIL
```
==> Check ASLR setting
Detected Slurm cluster. Checking ASLR on 6 nodes...

[!] ERROR: ASLR is enabled on Slurm node [bohr5] (value: 2).
    Please run: srun -w bohr5 sudo sh -c 'echo 0 > /proc/sys/kernel/randomize_va_space'
```

5surim (approved):
LGTM

PR #377, opened 2026-04-08
Support workload list for simulation

5surim:
We can selectively run workloads within a subsuite with this support instead of introducing all separate simulation fields for all workloads.
e.g. ["kafka", "xgboost", "cassandra"]


hlitz (commented):
Can we  eliminate the backslashes ? As in visualize?

5surim:
yes I also saw you removed them from visualize.
I just wonder if that is a valid JSON format. JSON identifies a field with double quotes if we just use double quotes within the double quotes.
I think we ignore this description field anyway when we parse the JSON, but it could include an incomplete field without backslashes with this double quote.

5surim:
So just removing backslashes did not work as they are double quotes inside double quotes that break the JSON parsing.
I replaced them with single quotes, but in the simulation fields, we might still need to use double quotes.

hlitz (approved):
LGTM

PR #380, opened 2026-04-13
Remove deprecated fuzzy option for gdown

hlitz:
The workflow fails because the option is not longer available. It should no longer be needed


5surim (approved):
LGTM

PR #386, opened 2026-04-26
Add dhrystone as a workload

### LINKED ISSUE #381: [FEATURE] Extend CI coverage
Ensure we test at lease one simpoint for memtrace, pt trace, and execution driven modes.

For execution driven, reduce simulation and warmup cycles to reduce CI usage.

FlyingFish800:
Rebase of @FlyingFish800's #386 onto current main, reduced to what it is: a new
workload. One file, `workloads/workloads_db.json`, +30 lines.

## Why

Every workload with an exec entry needs either a licensed benchmark install
bind-mounted from a lab host (183 of the 245 are SPEC) or an image the tests do
not build. So exec-driven simulation cannot be run on a machine that has only
this repository — which is also why CI has never covered that frontend.

Dhrystone is small enough to fix that: a single self-contained program, no input
data, nothing to license.

## Nothing special about it

- **One simpoint** (`cluster_id 0`), so it expands, runs and collects exactly
  like a simpointed SPEC run — same
  `<config>/<suite>/<subsuite>/<workload>/<cid>/` hierarchy, no special cases in
  the runners or in stat collection.
- **Its program comes from the descriptor's `application_dir`**, like SPEC.
  Whoever runs it installs it there; the workload entry says where the program
  lives, not where to get it.

## Verified

With the program in a local application directory and nothing else changed —
stock `sci`, stock runners, stock `scarab_stats.py`:

- the run lands in `.../dhrystone/dhrystone/dhrystone/0/` and finishes at
  `insts:10000006`, nine stat CSVs,
- `collected_stats.csv` gains the column
  `this_pr dhrystone/dhrystone/dhrystone 0`.

## Not here

Using this workload in CI is litz-lab/scarab#598: that job declares the
exec-driven simulation and installs the program on the runner. This repository
does not know about it.

Also dropped from the original PR: the `json/top_simpoint.json` →
`json/CI.json` rename, which would have broken Scarab's CI on merge
(`compile-scarab.yml` asks for `sci_id: top_simpoint` and reads it from
scarab-infra main), and the PT simulation, now #450.


json/CI.json
```diff
@@ -28,6 +28,22 @@
...
+      "simulation_type":"pt",
+      "warmup":10000000
+    },
+    {
+      "suite":"spec2017",
+      "subsuite":"rate_fp_v2",
+      "workload":"bwaves_r",
+      "cluster_id":0,
+      "simulation_type":"exec",
+      "warmup":1000000
```
5surim, line 46: Is it a typo or do we want to even reduce the warmup to 1M?

json/top_simpoint.json
```diff
@@ -28,6 +28,22 @@
       "cluster_id":109678,
       "simulation_type":"memtrace",
       "warmup":10000000
+    },
+    {
+      "suite":"datacenter",
+      "subsuite":"datacenter",
+      "workload":"clang",
```
5surim: Can you run cassandra instead of clang?

5surim (commented):
Hi @FlyingFish800 , thanks for the patch.
It seems CI fails, please have a look.

json/CI.json
FlyingFish800: This is not a typo, I remember in the lab meeting @hlitz said we want to reduce warmup to reduce CI usage. I noted this in issue #381

hlitz: The 1M reduction only applies to the CI. We do not want to reduce warmpu for Top3 simpoints! 

It s confusing as we have the concept of top simpoints. Can we rename the file to CI.json

workloads/workloads_top_simp.json
```diff
@@ -1,6 +1,7 @@
 {
   "google":{
     "google":{
+      "_scarab_sim_githash":"fdf3c92",
```
hlitz: what is this again?

workloads/workloads_top_simp.json
```diff
@@ -16,17 +17,26 @@
           {
             "cluster_id":1013732,
             "segment_id":0,
-            "weight":0.3333333333333333
+            "weight":0.3333333333333333,
```
hlitz: This does not seem to be related to the CI fix. Can you pull the workload_db mem edits into a separate PR. then we an just merge it in without me having to review 10000s of lines

hlitz (changes requested):

workloads/workloads_top_simp.json
FlyingFish800: For the 1M reduction: I could set the warmup to 10M and keep the filename the same, so users can still run the top simpoint test on their setup. We already modify this script a in the e2e workflow (~line 40), so I can move the 1M instruction reduction there

For the _scarab_sim_githash, this is the version of scarab for which the memory usage data was collected

hlitz (approved):
LGTM

PR #387, opened 2026-05-01
Fix job submission scalability issue

estodi:
Previously, it scanned the contents of all log files for every single job to find stale logs, resulting in extremely redundant file reads. Likewise, it also swept the slurm queue for every single job. Furthermore, it checked memory allocation info for all jobs, even when a job can be skipped and that info is not used. As a result, for example, when an experiment had thousands of completed logs, it needed to spend one hour just to submit 300 jobs. This PR addresses this scalability issue by consolidating file reads and checks across all jobs into single, global operations.

### DIFF
--- a/scripts/local_runner.py
+++ b/scripts/local_runner.py
@@ -24,6 +24,7 @@ from .utilities import (
         get_weight_by_cluster_id,
         image_exist,
         check_can_skip,
+        get_old_job_logs,
         remove_old_job_logs,
         print_simulation_status_summary,
         normalize_simulations,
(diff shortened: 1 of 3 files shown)


hlitz:
@estodi do I understand correctly that this PR does not change which log files/slurm queues are read by batching the these reads into fewer combined reads?

@FlyingFish800 Please have a look at @estodi patch which addresses the slow submission time we have all been observing. How does this patch affect the refactoring you are planning on the log files/slurm queue? 

This has high priority as submission times have become quite annoying

estodi:
@hlitz yes you are right. 

In the previous implementation, a job was submitted in the following manner.
1. Retrieve queue information.
2. If the job exists in the queue, skip it.
3. Retrieve old log files.
4. If old log files contain the job, delete them.
5. Submit the job.

Instead of repeating these steps for each job, this patch runs steps 1 and 3 once at the beginning and step 5 once at the end. Steps 2 and 4 still run for each job in this patch.

FlyingFish800:
@hlitz This should not significantly effect my refactoring

I have reviewed the code and it looks good to me. I am running a test now just to confirm it works and I will approve

FlyingFish800 (approved):
Tested and confirmed it works. Code LGTM!

PR #388, opened 2026-05-02
Fix --status crash on list workloads

5surim:
- get_simulation_job_identifiers in scripts/utilities.py iterated descriptor_data["simulations"] directly and crashed with TypeError: unhashable type: 'list' when a descriptor's workload field was a list (e.g., a sweep across multiple
  workloads in a single simulation entry).
  - Other callers — validate_simulation, get_simulation_jobs, get_image_list, local_runner, slurm_runner — already call normalize_simulations() to expand list-workload entries into individual entries. get_simulation_job_identifiers was
  the only outlier. Applying the same normalization here makes --status consistent with --sim.


hlitz (approved):
Lgtm

PR #389, opened 2026-05-07
Changed spec exec workloads to use allbench image

FlyingFish800:
Uses the allbench_traces docker image for execution-driven spec workloads. This will allow the git CI tests to work on images which we already automatically generate. Spec workloads are mounted into the docker container, so we don't need to compile them into the image anyways.

### Testing
Execution driven spec workloads with this commit work as well as they do with the spec image in the most recent version of scarab infra. 
There is an unrelated issue with some exec-driven spec workloads I have not been able to run successfully on the most recent scarab-infra (using spec image), but I am working with Moon to resolve this issue.

### DIFF
--- a/workloads/allbench_traces/Dockerfile
+++ b/workloads/allbench_traces/Dockerfile
@@ -5,5 +5,11 @@ FROM ubuntu:focal
 
 INCLUDE+ ./common/Dockerfile.common
 
+# For running Spec 2017 workloads
+RUN DEBIAN_FRONTEND=noninteractive apt-get install -y \
+    gfortran \
+    lsb-release \
+    software-properties-common
+
 # Start your application
 CMD ["/bin/bash"]
--- a/workloads/workloads_db.json
+++ b/workloads/workloads_db.json
@@ -1139,7 +1139,7 @@
         "simulation":{
           "prioritized_mode":"memtrace",
           "exec":{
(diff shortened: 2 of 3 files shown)


5surim:
Hi @FlyingFish800 

Thanks for working on this. I wanted to flag a concern with the approach for exec-driven SPEC support.
Each suite (SPEC, DCPerf, etc.) needs its own system environment for its binaries to run, and trying to combine them tends to surface library and OS-version conflicts. That's the reasoning behind keeping a separate Docker image per suite.
For allbench_traces specifically, the situation is simpler: since we're running pre-collected traces, we only need the Scarab setup (no application-specific dependencies). Adding more libraries here would just grow extra dependencies that would just raise the conflict risk.
Could we keep `allbench_traces` minimal and handle exec-driven SPEC in its own image instead?

Thanks,
Surim

FlyingFish800:
Hi @5surim, thank you for the feedback. The only change needed for running spec workloads (via a mounted applications directory) is 3 additional packages: gfortran, lsb-release, and software-properties-common. Does this still create a conflict risk concern? If so, I can switch back to a separate spec image

5surim:
Eventhough the change is small, this is not the correct and generic design that is applicable to all the different exec-driven simulations across all the different suites.

PR #394, opened 2026-06-03
DynamiRIO Mutithread-safe libfpg.so

5surim:
## Summary

Two fixes for multi-threaded Python workloads (numpy / torch / tokenizer worker pools) hitting the trace pipeline. No behaviour change for single-threaded workloads beyond switching the I/O backend in `libfpg.so`.

### `fpg.cpp` MT-safety and stack-overflow fixes

1. **SIGSEGV / file corruption from concurrent `std::cout` / `std::ofstream`.** DR clients share a single libc `FILE*` table, so two worker threads formatting output at the same time can race inside `fputc` and crash or interleave bytes. Replace every `std::cout` / `std::ofstream` with `dr_printf` / `dr_open_file` / `dr_write_file` (DR's own thread-safe I/O). The pcmap dump now buffers into a heap `std::string` and writes once instead of streaming line-by-line.
2. **Stack overflow inside `output_fingerprint`.** The function declared the fingerprint map and filename as by-value parameters. DR client threads have a ~56K stack and copying a `std::map` with tens of thousands of unique bb_ids (observed on CPython 3.10+ workloads) overflows it. Take both by `const&`.
3. **`DR_ASSERT(!for_trace)` at `fpg.cpp:734` crashed `libfpg.so`** when a thread spawned mid-trace from a worker pool hit a basic block not seen during fingerprinting. Remove the assertion and fall through to the existing fresh-bb_id allocation path.

### `Dockerfile.common`

1. Point the DR clone at `5surim/dynamorio` release_10.0.0, a fork of `moonseoppkim/dynamorio` with a `d_r_strcmp` fix that is hit by every multi-threaded Python workload during the trace phase.
2. On Ubuntu jammy (libc6-dev for glibc 2.34+), `/usr/lib/x86_64-linux-gnu/libpthread.so` is no longer shipped because pthread is merged into libc. DynamoRIO's CMake still calls `find_library(pthread)`, bakes the absolute path into make targets, and the scarab build later trips over the missing file. Restore the stub only when missing (older distros like Ubuntu 20.04 keep `libpthread.so` as a real symlink that must not be overwritten).
3. Add `time` to the apt set so per-stage wall-clock measurements work inside the container.

## Related

This is the substrate for #PR-TBD (trace-pipeline-hardening) which assumes the MT-safe `libfpg.so` and the new DR fork.

### DIFF
--- a/common/Dockerfile.common
+++ b/common/Dockerfile.common
@@ -25,9 +25,21 @@ RUN apt update && DEBIAN_FRONTEND=noninteractive apt install -y \
     gdb \
     graphviz \
     curl \
+    time \
     clang-tidy-18 \
  && update-alternatives --install /usr/bin/clang-tidy clang-tidy \
       /usr/bin/clang-tidy-18 100
+
+# Restore libpthread.so linker script on distros that omit it: Ubuntu jammy's
+# libc6-dev no longer ships /usr/lib/x86_64-linux-gnu/libpthread.so because
+# glibc 2.34+ merged libpthread into libc, but DynamoRIO's CMake bakes
+# find_library(pthread) absolute paths into make targets that later break
+# scarab build. Only create the stub when the file is missing; on older
+# distros (e.g. Ubuntu 20.04) libpthread.so is a symlink to the runtime
+# library that must not be overwritten.
+RUN if [ ! -e /usr/lib/x86_64-linux-gnu/libpthread.so ]; then \
+        echo 'GROUP ( /lib/x86_64-linux-gnu/libpthread.so.0 )' > /usr/lib/x86_64-linux-gnu/libpthread.so; \
+    fi
 
 RUN curl -fsSL https://deb.nodesource.com/setup_22.x | bash -
 RUN apt install -y nodejs
@@ -59,7 +71,7 @@ WORKDIR $tmpdir
 # We will switch back to the official DynamoRIO once the fetched instruction counting logic is resolved (https://github.com/DynamoRIO/dynamorio/issues/4948)
 # RUN cd $tmpdir && wget https://github.com/DynamoRIO/dynamorio/releases/download/release_10.0.0/DynamoRIO-Linux-10.0.0.tar.gz && tar -xzvf DynamoRIO-Linux-10.0.0.tar.gz
 RUN cd $tmpdir && \
-    git clone --branch release_10.0.0 --depth 1 https://github.com/moonseoppkim/dynamorio.git DynamoRIO-Linux-10.0.0 && \
+    git clone --branch release_10.0.0 --depth 1 https://github.com/5surim/dynamorio.git DynamoRIO-Linux-10.0.0 && \
     cd DynamoRIO-Linux-10.0.0 && \
     git submodule update --init --recursive && \
     cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo \
--- a/fingerprint_src/fpg.cpp
+++ b/fingerprint_src/fpg.cpp
@@ -2,11 +2,11 @@
 #include "drmgr.h"
 #include "drutil.h"
 #include "droption.h"
-#include <iostream>
 #include <vector>
 #include <map>
-#include <fstream>
 #include <string>
+#include <cstring>
+#include <cstdio>
 #ifdef WINDOWS
 # define DISPLAY_STRING(msg) dr_messagebox(msg)
 #else
@@ -149,7 +149,10 @@ clean_call(uint instruction_count, uint64 bb_id, uint64 segment_size, uint emula
 
 // identical to memtrace post processing
 // excepty for assertions
-uint64_t output_fingerprint(std::string file_name, std::map<uint64_t, uint64_t> fingerprint, bool exit);
+// NOTE: fingerprint is passed by const reference. Passing by value copies the
+// entire std::map on the DR client stack (which is only 56K by default) and
+// overflows on workloads with large unique bb_id counts like CPython 3.10+.
+uint64_t output_fingerprint(const std::string &file_name, const std::map<uint64_t, uint64_t> &fingerprint, bool exit);
 void output_total_instrs_in_seg(per_thread_data *t_data, uint64 seg_idx);
 
 DR_EXPORT void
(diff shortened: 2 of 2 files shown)


common/Dockerfile.common
```diff
@@ -59,7 +71,7 @@ WORKDIR $tmpdir
 # We will switch back to the official DynamoRIO once the fetched instruction counting logic is resolved (https://github.com/DynamoRIO/dynamorio/issues/4948)
 # RUN cd $tmpdir && wget https://github.com/DynamoRIO/dynamorio/releases/download/release_10.0.0/DynamoRIO-Linux-10.0.0.tar.gz && tar -xzvf DynamoRIO-Linux-10.0.0.tar.gz
 RUN cd $tmpdir && \
-    git clone --branch release_10.0.0 --depth 1 https://github.com/moonseoppkim/dynamorio.git DynamoRIO-Linux-10.0.0 && \
+    git clone --branch release_10.0.0 --depth 1 https://github.com/5surim/dynamorio.git DynamoRIO-Linux-10.0.0 && \
```
hlitz, line 74: I did not get this change. We want to keep Moon's patch. How does your fork differ from Moon and upstream?

hlitz (commented):
@moonseoppkim can you have a look at this PR. Generally looks good to me.

common/Dockerfile.common
moonseoppkim: I don't agree to replace that url according to these previous PRs. (https://github.com/litz-lab/scarab-infra/pull/279, https://github.com/DynamoRIO/dynamorio/pull/7772)

I had introduced a new feature(count_fetched_instrs) and adjusted a method to count instruction(regarding REP) for scarab-infra.

5surim: The new url is a fork of Moon's Dynamorio fork. So it includes all Moon's patches. Mine includes multi-thread-related fix on top of it.

moonseoppkim: Unfortuantly, I can't find any related PRs on your fork.
(https://github.com/5surim/dynamorio)

5surim: https://github.com/5surim/dynamorio/tree/release_10.0.0

It is checking out to release_10.0.0 (Please look at the command)

moonseoppkim: Whoops, sorry

let me check again

moonseoppkim: @5surim, if you don't mind, could you give me write access to your fork? 
By any chance, I need to push some changes on top of your work once you left.

5surim: Thanks, I just invited

fingerprint_src/fpg.cpp
```diff
@@ -739,33 +770,61 @@ event_app_analysis(void *drcontext, void *tag, instrlist_t *bb,
...
+
+            // inspections
+            if (!for_trace && !translating) {
+                t_data->inspect_case_as_built[NON_TRACE_NON_TRANSLATING_RECREATION]++;
             }
-            dr_printf("existed bb:\n");
-            for (app_pc pc : existed_bb_pc) {
-                dr_printf("%llu\n", (uint64)pc);
+        } else {
+            // Same first_addr but different content. This is legal on Python
```
moonseoppkim, line 792: @5surim, Could you provide more details(logs) for this case?

I've never seen this case before with spec workloads. So, I'm a little bit curious about the practical flow of logs.

5surim: You're right that SPEC won't fire this, and the reason is structural to SPEC rather than something I patched around — so your observation is accurate. SPEC C/C++ workloads under tracing are essentially:

- Single-threaded per copy (`intrate`/`fprate` is multi-process; OpenMP variants are usually pinned with `OMP_NUM_THREADS=1` for deterministic SimPoint clustering).
- Static binaries with no JIT and no `dlopen`/`dlclose` after startup.
- Deterministic instruction count for fixed reference inputs.

DR's cached translation for a basic block at a given `first_addr` therefore stays valid for the entire run — content never changes underneath us, so the `addr_to_bb_for_trace` lookup either matches exactly (normal path) or misses entirely (new bb), never "same addr, different content." The assert never fires.

What does fire it: workloads where the code at a given PC gets *rewritten in place* after DR has translated it. The three triggers I've hit while bringing up Python-based workloads in a downstream fork:

1. **CPython 3.10+ PEP 659 adaptive interpreter.** The bytecode-specialising rewriter mutates inline-cache slots (`LOAD_ATTR` → `LOAD_ATTR_INSTANCE_VALUE`, `CALL` → `CALL_PY_EXACT_ARGS`, etc.) at the same PC as the interpreter warms up. DR sees a BB at the same `first_addr` with different content. Hits almost any non-trivial Python workload after a few thousand bytecode dispatches.
2. **JIT'd trampolines.** PyPy, Numba, and V8 native-code patching (Node.js CLIs) repatch callsites after warmup. Short-lived `mprotect(RWX)` regions inside dlopen'd .so plt/got rewrites also count.
3. **DR code-cache flushes** can re-build a BB whose upstream code is still mutating. Relatively rare on its own; combines with 1 / 2.

Reproduction is any CPython 3.10+ workload traced through `cluster_then_trace`. Once a Python workload is registered upstream (TBD), this should be straightforward to hit; in the meantime the trigger is structurally guaranteed by PEP 659 on CPython 3.10+. The stderr line emitted by the new path looks like
```
[140234567890123] bb at first_addr 94276533751536 re-materialized with different content (instr count 3 -> 4, bb size 12 -> 16); re-keying as new bb_id
```
On a typical Python workload I see this fire on the order of 10–50 times during fingerprinting, all in the PEP 659 inline-cache regions of CPython's `_PyEval_EvalFrameDefault`.

Before the change, hitting this path called `DR_ASSERT(...)` inside libfpg.so and SIGSEGV'd the fingerprint job mid-run. Now it allocates a fresh `bb_id`, overwrites the `addr -> bb_id` mapping for that PC, and lets the trace continue.

Happy to attach a sample stderr dump from a real run if it would help.

moonseoppkim: Thanks for your kind response!

moonseoppkim:
LGTM

moonseoppkim (approved):

PR #395, opened 2026-06-03
Trace pipeline hardening: MT-safe drrun, per-segment isolation, opt-in empty-segment recovery

5surim:
## Summary

Workload-agnostic refinements accumulated while running multi-threaded Python workloads (numpy / torch / tokenizer worker pools) through `cluster_then_trace`. **Chained on #394** — assumes the MT-safe `libfpg.so` and the new DR fork land first.

This PR contains **only fixes to existing modes 1 / 2 / 3** (`cluster_then_trace`, `trace_then_cluster`, `iterative`). The new per-segment Slurm fan-out feature (modes 4 / 5 plus outer wiring) is split into its own follow-up PR for clearer review scope.

### `run_simpoint_trace.py` changes

**DynamoRIO command hardening**
- Add `-disable_rseq` to every `drrun` (fingerprint + all trace modes). Without it, `drrun` crashes on glibc `rseq()` when the workload spawns threads after DR attach.
- Make `-ipc_name` per-segment in `cluster_then_trace`. Observed failure: two parallel trace variants land on the same node; drcachesim's default ipc_name (`drcachesimpipe`) shares per-host `/tmp` coordination state, and one `drrun` wrote its trace into the other variant's outdir. Per-segment ipc_name isolates them.

**Per-workload DR config isolation**
- Override `HOME` to the workload home for `drrun` invocations instead of `DYNAMORIO_CONFIGDIR`. The latter is ignored when `drrun` is invoked via shell rather than directly; `HOME` override reliably scopes DR's `~/.dynamorio` cache per workload and prevents cross-contamination between concurrent workloads on the same host.
- Clean stale fingerprint outputs (`bbfp.*`, `pcmap.*`) before retry.

**Fingerprint retry + determinism**
- Wrap the fingerprint `drrun` in a 3-attempt retry loop with stderr tail on failure; the first attempt occasionally fails due to a transient DR initialisation race on hot containers.
- Add a `PYTHONHASHSEED` + OMP/MKL/OPENBLAS/NUMEXPR/TOKENIZERS thread-limit prefix so fingerprinting is single-threaded and deterministic.

**Multi-threaded workload handling**
- `minimize_simpoint_traces` now tolerates multiple `trace.zip` files per segment (one per thread): pick the largest and delete the rest.
- Clustering picks the bbfp file with the most segments when `fpg.cpp` emits one per thread.

**Empty-segment handling: opt-in auto-removal (revised design)**
- Default is **fail-hard** when raw2trace produces no usable trace for a segment, so the user notices when the workload's instruction count is non-deterministic.
- Opt-in recovery via env var `EMPTY_SEGMENT_ACTION=remove` (set on the `trace_configuration`'s `env_vars`). When enabled, segments with simpoint weight below 1% are removed from `opt.p.lpt0.99` / `opt.w.lpt0.99` (with `.bak` backups) and the remaining weights renormalised. Above 1%, still fail hard.
- Rationale: silently dropping segments biases the workload representation toward earlier phases, so the conservative default surfaces the problem and lets the user decide (rerun, switch to `trace_then_cluster`, fix workload determinism, or opt into removal).

**Concurrency**
- `DR_JOBS` env var (auto-detect cpu_count clamped to [2,40] if unset) replaces hard-coded `-jobs 40` in every `drrun` / `drraw2trace` site.

**Other**
- `portabilize_trace.py` is skipped per-segment when `bin/` already has more than 10 `.so` files (resumed segments).
- Skip-if-already-minimized check at the start of `minimize_simpoint_traces` avoids redoing finished segments on rerun.

### `docs/README.trace.md`

Replace the original step-by-step "how to add DCPerf" tutorial with a structured reference for the actual pipeline: descriptor schema, the three `trace_type` modes, the 8 pipeline stages with per-stage resume semantics, memory-management model, multi-threaded Python workload handling, env-var reference, on-disk layout, and a troubleshooting section. Documents the new `EMPTY_SEGMENT_ACTION` opt-in and warns about the bias of silent segment removal.


hlitz (commented):
lgtm but I'd also like @moonseoppkim to have a quick look as he is familiar with the code

common/scripts/run_simpoint_trace.py
```diff
@@ -76,16 +161,48 @@ def minimize_simpoint_traces(cluster_map, workload_home, warmup_chunks):
...
+            dest_zip = os.path.join(dest_trace_dir, f"{segment_id}.zip")
+            if os.path.isfile(dest_zip) and os.path.getsize(dest_zip) > 0:
+                continue
             trace_dir = os.path.join(workload_home, "traces_simp", str(segment_id))
             trace_files = subprocess.getoutput(f"find {trace_dir} -name 'dr*.trace.zip' | grep 'drmemtrace.*.trace.zip'").splitlines()
+            trace_files = [f for f in trace_files if f.strip()]
             num_traces = len(trace_files)
-            if num_traces != 1:
+            if num_traces == 0:
+                # raw2trace ran but produced no usable trace (e.g. segment past
```
moonseoppkim, line 188: I might have shared last time that I've experienced a similar issue with draw2trace(minizip bug at Dynamorio) from some spec workloads.
 
In particular, It happened when I tried to translate a huge raw file into trace.
(* Output trace folder was empty as mini-zip bug)

So, this was my work-around code for that.
 
```
drraw2trace with lz4(not zip) -> raw lz4 -> zip
 
/dynamorio_full_mine/dynamorio/tools/bin64/drraw2trace \
    -verbose 1 \
    -indir /vvvv/403.gcc_test/gcc/traces_simp/8766/raw \
    -alt_module_dir /vvvv/403.gcc_test/gcc/traces_simp/8766/bin \
    -chunk_instr_count 10000000 \
    -jobs 1 \
    -compress lz4 -> (default : zip -> using mini-zip from dynamorio as default)


 /vvvv/403.gcc_test/gcc/traces_simp/8766/trace/window.0000/drmemtrace.gcc_base.gcc43-64bit.00123.6516.trace.lz4
 

zip 8766.zip /vvvv/403.gcc_test/gcc/traces_simp/8766/trace/window.0000/drmemtrace.gcc_base.gcc43-64bit.00123.6516.trace
```

I can prepare this change later, 
but I would like to put very brief explanation about this phenomenon at the below prints.

common/scripts/run_simpoint_trace.py
```diff
@@ -256,52 +374,134 @@ def cluster_then_trace(workload, suite, simpoint_home, bincmd, client_bincmd, si
...
+            # Disable restartable sequences — DR doesn't fully support rseq on
+            # newer glibc, causing "entries are not in a loaded segment" errors.
+            drio_extra += " -disable_rseq"
+            # Force single-threaded execution during fingerprinting to avoid
+            # DynamoRIO crashes from concurrent module loads in multi-threaded
+            # Python workloads (torch/faiss/sentence-transformers). The crash
+            # is in DR's d_r_strcmp called with a corrupted module-name pointer
+            # when multiple threads dlopen/dlclose simultaneously. Single-threaded
+            # fingerprinting is also more deterministic for SimPoint.
+            thread_limit_prefix = "PYTHONHASHSEED=0 OMP_NUM_THREADS=1 MKL_NUM_THREADS=1 OPENBLAS_NUM_THREADS=1 NUMEXPR_NUM_THREADS=1 TOKENIZERS_PARALLELISM=false"
```
moonseoppkim: Do you think is it okay to use this envs only for finger_print_generator? 

How about binary_command execution?
```
        max_parallel = int(os.environ.get("TRACE_PARALLEL", "1"))
            batch = trace_cmds[batch_start:batch_start + max_parallel]
```

5surim:
Pushed `e2114ba` addressing both review points. Summary:

**#395:406 — single-thread env vars on trace-phase drrun, not just fingerprinting.**
Promoted the prefix to a module-level `THREAD_LIMIT_ENV_PREFIX` constant and applied it to every `drrun` in the trace phase across all three modes:
- `cluster_then_trace` (mode 1) — per-segment `trace_cmd` in the segment loop (the exact site you flagged)
- `trace_then_cluster` (mode 2) — whole-app `trace_cmd`
- `iterative_trace` (mode 3) — per-timestep `trace_cmd`

For modes 2 and 3 the existing `subprocess.run(shlex.split(...))` won't honour leading `KEY=VALUE` assignments (the kernel would treat them as the program name), so those sites are now wrapped in `/bin/sh -c`. Also propagated to mode 5 (`trace_single_segment`) on #398 in the rebased branch so per-segment Slurm fan-out gets the same treatment.

**#395:173 — expanded empty-trace error message.**
The pipeline now prints both failure modes inline with concrete remediation:
1. Non-deterministic workload — points users at `trace_then_cluster` or the opt-in `EMPTY_SEGMENT_ACTION=remove`.
2. **DR minizip bug on large raw** (your case) — spells out the `drraw2trace -compress lz4` workaround with the exact command. Happy to land your full lz4→manual-zip change in a follow-up PR once it's ready.

Note on SPEC blind spots from your other comment on #394: agree the threading-driven items here (PEP 659 fpg path, multi-thread bbfp, `THREAD_LIMIT_ENV_PREFIX`, MT-aware raw selection) don't fire on SPEC because each copy is single-threaded with a static binary — they were all driven by Python / agent workloads. The minizip path is the one issue that's size-driven not thread-driven, which is why you've independently hit it on SPEC.

5surim:
@moonseoppkim , friendly ping on this. The two review points were addressed already. Could you take another look when you get a chance?
This PR is currently blocking #397 and #398.

Thanks!
Surim

common/scripts/run_simpoint_trace.py
```diff
@@ -66,6 +88,79 @@ def get_cluster_map(workload_home):
 
     return cluster_map
 
+EMPTY_TRACE_WEIGHT_THRESHOLD = 0.01  # auto-remove segments with weight < 1%
```
moonseoppkim: Why do we need this feature even though we have same feature at simpoint?

```
-coveragePct .99 << simpoint feature

spCmd="$tmpdir/simpoint -maxK $maxK -dim 100 -fixedLength off -numInitSeeds 10 -loadFVFile $FPFILE -saveSimpoints $OUTDIR/simpoints/opt.p -saveSimpointWeights $OUTDIR/simpoints/opt.w -saveVectorWeights $OUTDIR/simpoints/vector.w -saveLabels $OUTDIR/simpoints/opt.l -coveragePct .99 &> $OUTDIR/simpoints/simp.opt.log"
```

common/scripts/run_simpoint_trace.py
```diff
@@ -256,52 +405,133 @@ def cluster_then_trace(workload, suite, simpoint_home, bincmd, client_bincmd, si
...
+            print("generate fingerprint.. skipped (already exists)")
+        else:
+            print("generate fingerprint..")
+            if client_bincmd:
+                subprocess.Popen("exec " + client_bincmd, stdout=subprocess.PIPE, shell=True)
+            start_time = time.perf_counter()
+            drio_extra = drio_args if drio_args else ""
+            # Disable restartable sequences — DR doesn't fully support rseq on
+            # newer glibc, causing "entries are not in a loaded segment" errors.
+            drio_extra += " -disable_rseq"
```
moonseoppkim, line 435: Could you confirm that you've tested this latest code(adding disable_resq option) with a single-threaded workload? (etc. SPEC)

common/scripts/run_simpoint_trace.py
```diff
@@ -256,52 +405,133 @@ def cluster_then_trace(workload, suite, simpoint_home, bincmd, client_bincmd, si
...
+            # Force single-threaded execution during fingerprinting to avoid
+            # DynamoRIO crashes from concurrent module loads in multi-threaded
+            # Python workloads (torch/faiss/sentence-transformers). The crash
+            # is in DR's d_r_strcmp called with a corrupted module-name pointer
+            # when multiple threads dlopen/dlclose simultaneously. Single-threaded
+            # fingerprinting is also more deterministic for SimPoint.
+            fp_cmd = f"HOME={dr_home} {THREAD_LIMIT_ENV_PREFIX} {dynamorio_home}/bin64/drrun -max_bb_instrs 4095 -opt_cleancall 2 {drio_extra} -c $tmpdir/libfpg.so -no_use_bb_pc -segment_size {seg_size} -output {workload_home}/fingerprint/bbfp -pcmap_output {workload_home}/fingerprint/pcmap -- {bincmd}"
+            # DynamoRIO has a non-deterministic crash in d_r_strcmp when
```
moonseoppkim: Do you think this non-deterministic crash could be solved by customizing our own DynamoRIO build in terms of d_r_strcmp implementation?

docs/README.trace.md
```diff
@@ -1,192 +1,360 @@
...
+- `client_bincmd`: optional client command (for client-server workloads)
+- `trace_type`: `"cluster_then_trace"`, `"trace_then_cluster"`, or `"iterative_trace"`
+- `dynamorio_args`: extra DynamoRIO flags
+- `clustering_k`: override SimPoint cluster count (null = auto)
+- `slurm_options`: extra sbatch flags (ignored in manual mode)
+- `trace_mem_mb`: optional per-workload memory override (MB)
+
+## Trace Types
+
+### `cluster_then_trace` (recommended)
```
moonseoppkim, line 96: Worth noting for debugging: trace_then_cluster is much easier to debug, everything runs off one fixed trace, so any issue can be reproduced exactly. 

cluster_then_trace uses two separate executions, so failures may not reproduce.

common/scripts/run_simpoint_trace.py
5surim: These are different. SimPoint's `coveragePct .99` controls clustering. It picks K clusters that cover 99% of execution. `EMPTY_TRACE_WEIGHT_THRESHOLD` is a post-tracing safety net that removes segments whose actual traced output is empty or corrupt but whose SimPoint weight is negligible (<1%). It handles the case where `cluster_then_trace` produces anempty segment due to a DR bug. If the segment only accounts for <1% of execution, it's safe to drop it rather than fail the whole pipeline. SimPoint's coveragePct can't handle this because it runs before tracing.

common/scripts/run_simpoint_trace.py
5surim: `-disable_rseq` is a safe no-op on workloads that don't use rseq (restartable sequences). Single-threaded SPEC workloads won't use rseq, so the flag has no effect. But I plan to test with SPEC2026 and deliver fixes if there are any issues.

common/scripts/run_simpoint_trace.py
5surim: The crash is in DR's `d_r_strcmp` called with a corrupted module-name pointer during concurrent `dlopen/dlclose`. So yes.

common/scripts/run_simpoint_trace.py
5surim: Thanks for sharing the workaround. The minizip bug in DynamoRIO is a known issue with large raw files. Your lz4 approach (drraw2trace with `-compress lz4`, then repack to zip) is a solid workaround.

The `EMPTY_TRACE_WEIGHT_THRESHOLD` mechanism in this PR handles a related but different symptom: when a segment's trace output is empty (regardless of cause — minizip bug, OOM, DR crash), it removes that segment from the SimPoint set if its weight is negligible (<1%). This prevents the simulation from failing on missing traces.

I'd be happy to add a brief comment near the drraw2trace calls noting the minizip issue and the lz4 fallback option. Would you like to prepare that as a follow-up PR, or should I include it here?

common/scripts/run_simpoint_trace.py
5surim: Good question. `THREAD_LIMIT_ENV_PREFIX` (OMP_NUM_THREADS=1, etc.) is already applied to all drrun commands — fingerprinting (line 442), cluster tracing (lines 577/579), and whole-app tracing (line 269). So the workload itself is single-threaded in every phase.

`TRACE_PARALLEL` is a separate concern: it controls how many *segment traces* run concurrently (each segment launches a full drrun process). Since each drrun process can use 500MB–1GB, launching all segments in parallel (e.g. 30+) easily exceeds Slurm memory limits. The default of 1 is conservative; users can increase it with `TRACE_PARALLEL=4` if they have enough memory.

So the env-var scoping is intentional: thread-limiting applies to every drrun invocation, while `TRACE_PARALLEL` is specific to the cluster_then_trace batching loop.

docs/README.trace.md
5surim: Good suggestion. I added this as a debugging tip in the latest push (commit 11861d5). The note is now under the `trace_then_cluster` section in README.trace.md.

common/scripts/run_simpoint_trace.py
moonseoppkim: Thx!

common/scripts/run_simpoint_trace.py
moonseoppkim: Could you add a note in the code that this retry is a temporary workaround? 
The root cause should be addressed on the DynamoRIO side.

docs/README.trace.md
moonseoppkim: Thanks!

common/scripts/run_simpoint_trace.py
5surim: Done — updated the comment to say "TEMPORARY WORKAROUND" and note that the root cause should be fixed upstream in DynamoRIO (commit e86ccce).

moonseoppkim (approved):
LGTM

PR #396, opened 2026-06-03
Perf pipeline v2: automated top-down collection + container hardening

5surim:
## Summary

Turns `./sci --perf` into an automated collection pipeline (top-down L1, execution time, peak RSS) instead of an interactive shell, and hardens the supporting container entrypoints.

**Breaking change**: `./sci --perf <descriptor>` now triggers automated collection. The previous interactive-shell behaviour is preserved verbatim under the new `./sci --perf-interactive <descriptor>` flag. Discussed: no existing automated consumers of the old `--perf` behaviour.

Independent of #394 and #395 — different files, different scope. Can land in any order.

### `scripts/run_perf.py` (399 net new lines)

- **Adaptive repeat scaling**: target ~600s total measurement, `max(1, min(15, int(600 / single_run_seconds)))`. Replaces fixed-N which over-measured fast workloads and under-measured slow ones.
- **Peak RSS capture** during warmup via Python's `resource.RUSAGE_CHILDREN`. Persisted to `workloads_db.json` under `performance.peak_rss_mb` and consumed by `./sci --trace` for Slurm memory sizing.
- **Container bootstrap sentinel**: `/tmp_home/.scarab_perf_ready` ensures `root_entrypoint.sh` + `perf_entrypoint.sh` run exactly once per container lifetime. `sync_perf_support_files()` `docker cp`'s the latest scripts (`utilities.sh`, `root_entrypoint.sh`, `user_entrypoint.sh`, `perf_entrypoint.sh`, plus optional per-image `workload_*_entrypoint.sh`) into the container on each run so iterating on the host scripts does not require an image rebuild.
- **Top-down L1 + execution time + workloads_db.json output**: run `pmu-tools/toplev` with `-l1 --single-thread`, pin via `taskset -c 10`, parse Frontend/Backend/BadSpec/Retiring; measure execution time via `perf stat -C 10` with adaptive repeats.
- **Root-user hardening**: compute `HOME` based on user (`/tmp_home` for root, `/home/<user>` otherwise); pass `user_id` / `group_id` / `username` / `HOME` / `APP_GROUPNAME` / `APPNAME` env on `docker run` so the entrypoint runs cleanly as either root or a regular user.
- **Single-core pinning + thread-limit env**: all measurement commands pin to core 10 via taskset and inherit `PYTHONHASHSEED` + `OMP/MKL/OPENBLAS/NUMEXPR/TOKENIZERS` thread limits from the Docker image so multi-threaded workloads measure as single-threaded on the measured core.

### `common/scripts/perf_entrypoint.sh` (41 lines)

- Each `apt-get install` guarded by an `apt-cache` existence check so containers without `linux-tools-$(uname -r)` (common on minimal cloud images) don't fail the install step. Also probe `linux-cloud-tools-*` and `linux-perf` as alternates.
- After install, look for the perf binary under `/usr/lib/linux-tools` first, then fall back to PATH (`command -v perf`). Symlink to `/usr/local/bin/perf` only if found.
- pmu-tools clone is now idempotent: skipped if the `.git` directory or the `pmu-tools` dir already exists.
- `set -e` so any unguarded failure surfaces.

### \`common/scripts/root_entrypoint.sh\` (22 lines)

- `groupadd` / `useradd` are now guarded by `id` / `getent` checks so the entrypoint can be re-run on a container where the user is already set up.
- Skip the user-creation block entirely when `username` / `user_id` / `group_id` env vars are not set (root-user descriptors).
- `chmod` on `libdynamorio.so` guarded by an existence check so images built without DynamoRIO (perf-only descriptors) don't error.

### \`sci\` CLI

- `--perf <descriptor>` → action `collect` (was: action `launch`)
- New `--perf-interactive <descriptor>` → action `launch` (the previous `--perf` behaviour)
- `dtype == \"perf\"` guard updated to accept both `launch` and `collect`.

### `docs/README.perf.md`

Replace the original step-by-step tutorial (\"install perf, run feedsim, here's the top-down output\") with a structured reference: descriptor schema, 6-stage pipeline, results storage schema, multi-threaded guidance, container-reuse behaviour. DCPerf `fibers_benchmark` from `json/perf.json` used as the inline example.

### DIFF
--- a/common/scripts/perf_entrypoint.sh
+++ b/common/scripts/perf_entrypoint.sh
@@ -1,8 +1,39 @@
 #!/bin/bash
 #set -x #echo on
 
-cd $tmpdir
-apt-get install -y linux-tools-common linux-tools-generic linux-tools-`uname -r`
-sudo ln -s "$(find /usr/lib/linux-tools/*/perf | head -1)" /usr/local/bin/perf
-alias perf=$(find /usr/lib/linux-tools/*/perf | head -1)
-cd $tmpdir && git clone https://github.com/andikleen/pmu-tools.git
+set -e
+
+cd "${tmpdir:-/tmp_home}"
+export DEBIAN_FRONTEND=noninteractive
+
+apt_package_exists() {
+  apt-cache show "$1" >/dev/null 2>&1
+}
+
+install_if_available() {
+  if apt_package_exists "$1"; then
+    apt-get install -y "$1"
+  fi
+}
+
+apt-get update
+install_if_available linux-tools-common
+install_if_available linux-tools-generic
+install_if_available "linux-tools-$(uname -r)"
+install_if_available "linux-cloud-tools-$(uname -r)"
+install_if_available linux-cloud-tools-generic
+install_if_available linux-perf
+
+perf_bin="$(find /usr/lib/linux-tools /usr/lib -path '*/perf' -type f 2>/dev/null | grep 'linux-tools' | head -1)"
+if [ -z "$perf_bin" ] && command -v perf >/dev/null 2>&1; then
+  perf_bin="$(command -v perf)"
+fi
+if [ -n "$perf_bin" ]; then
+  ln -sf "$perf_bin" /usr/local/bin/perf
+fi
+
+if [ ! -d pmu-tools/.git ]; then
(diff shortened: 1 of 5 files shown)


hlitz (approved):
LGTM

PR #398, opened 2026-06-03
Trace pipeline: per-segment Slurm fan-out (parallel_segments)

5surim:
## Summary

Per-segment Slurm fan-out for the trace pipeline. **Chained on #395** — assumes the MT-safe drrun flags, per-segment ipc isolation, and DR config isolation from the trace-pipeline-hardening PR land first.

For workloads with many simpoints (50+), the default single-job trace pipeline serially traces one segment after another inside one Slurm allocation, often running for 10+ hours. Setting `"parallel_segments": true` in the descriptor activates a 3-phase Slurm pipeline that runs each segment as an independent job.

```
Phase 1 (1 job):   fingerprint + cluster (mode 4)
    │             produces opt.p.lpt0.99, opt.w.lpt0.99
    ▼ (Phase 1 sbatch script reads opt.p.lpt0.99 and submits Phase 2 jobs)
Phase 2 (N jobs):  trace + raw2trace + minimize one segment each (mode 5)
    │
    ▼ (--dependency=afterany:<all Phase 2 job ids>)
Phase 3 (1 job):   finish_trace (consolidates traces + updates workloads_db.json)
```

Only valid with `workload_manager: "slurm"` and `trace_type: "cluster_then_trace"`. The descriptor validator rejects mixed configurations.

### `common/scripts/run_simpoint_trace.py`

Re-introduces the two pipeline modes that were stripped from #395 to keep that PR's scope to fixes:

- **Mode 4 (`cluster_only`)** — fingerprint + clustering only, no tracing. Inherits all the hardening from #395 (PYTHONHASHSEED + thread-limit prefix, 3-attempt fingerprint retry, HOME-based DR config isolation, MT-aware bbfp selection).
- **Mode 5 (`trace_single_segment`)** — traces, portabilizes, raw2traces and minimises a single `(segment_id, cluster_id)`. Per-segment `-ipc_name` isolates concurrent drrun instances; MT-aware thread filter on `raw/window.0000` (keep largest raw file, drop helper threads) before raw2trace. Calls `minimize_simpoint_traces` with a single-entry cluster_map so the opt-in empty-segment semantics from #395 apply per-segment too.
- Both modes are dispatched on `simpoint_mode == "4"` / `"5"`; mode 5 requires `--segment_id` and `--cluster_id`.

### `scripts/utilities.py`

- Extend `generate_single_trace_run_command` to translate `"cluster_only"` to mode 4 and `"trace_single_segment"` to mode 5, and to append `--segment_id` / `--cluster_id` when provided.
- Quote `drio_args` so dash-prefixed values (e.g. `"-disable_rseq"`) are passed as a single argparse value rather than being parsed as a flag.
- Extend `write_trace_docker_command_to_file` with optional `segment_id` / `cluster_id` kwargs.
- Add `write_phase2_sbatch_tail`. Appended to the Phase 1 sbatch, it reads `opt.p.lpt0.99` inside the Slurm job, skips already-complete segments by checking `traces_simp/trace/{segment_id}.zip`, submits one sbatch per remaining segment using a persisted Phase 2 template, and submits a Phase 3 `finish_trace` sbatch with `--dependency=afterany`. If every segment is already complete, submits finalisation directly without dependencies.

### `scripts/slurm_runner.py`

- Add `SEGMENT_MEM_MIN_MB`, `PHASE1_MEM_DEFAULT_MB`, `PHASE3_FINALIZE_MEM_MB` constants. Simple fixed defaults; peak_rss-based sizing from `workloads_db.json` is a planned follow-up.
- Extend `run_tracing` signature with `descriptor_path` (needed so the Phase 3 sbatch can re-invoke `python -m scripts.run_trace -d <descriptor> -f`).
- Inside `run_tracing`, add `_run_parallel_segments_trace` that writes a persisted Phase 2 template (under `{trace_dir}/{workload}/phase2_template.sh` so it survives host tmp-file cleanup) and the Phase 1 script with the Phase 2+3 sbatch tail appended, then submits Phase 1.
- Branch in the trace-configuration loop dispatches to `_run_parallel_segments_trace` when `parallel_segments` is true and `trace_type == cluster_then_trace`.
- Skip the host-side `finish_trace` at the end in parallel_segments mode; Phase 3 runs it as a Slurm job.

### `scripts/run_trace.py`

- Add `--finish-trace` / `-f` CLI flag for invoking `finish_trace` directly. Used by the Phase 3 sbatch generated by `write_phase2_sbatch_tail`.
- Pass `descriptor_path` through to `slurm_runner.run_tracing`.
- Add `parallel_segments` validation in `verify_descriptor`: only valid with `workload_manager="slurm"` and `trace_type="cluster_then_trace"`. Reject mixed descriptors with a clear error rather than letting Phase 1 fail at runtime.

### `docs/README.trace.md`

New "Parallel Segment Tracing" section: 3-phase diagram, eligibility constraints, per-segment memory defaults, resume semantics (skip already-minimised segments), descriptor field shape.

### What's deferred

- **--status / --kill integration with Phase 2 / Phase 3 jobs.** Phase 1 jobs already appear via the existing `{image}_{workload}_{trace}_cluster_only_{user}_tmp_run.sh` pattern; Phase 2 jobs invoke `phase2_template.sh` directly (no `_tmp_run.sh` suffix) and Phase 3 uses `sbatch --wrap` (no job name). For now, users running parallel_segments mode can use `squeue -u <user>` directly. Documented in README.
- **peak_rss_mb-based memory sizing** for Phase 1 / Phase 2 / Phase 3 jobs. Current defaults are fixed constants. A follow-up can wire `workloads_db.json` lookups on top without touching the surrounding pipeline.
- **OOM auto-scaling via sacct** (re-submit with bumped memory). Independent feature; same PR boundary.


hlitz (dismissed):
lgtm

moonseoppkim:
Hi, @5surim 

I assume you are running the tracing jobs on Slurm?

I've never traced on Slurm myself(only at local machine[mainly ohm]), because I wasn't sure the results would be identical across different nodes. If the fingerprinting/clustering job and the per-segment tracing jobs land on nodes with different CPU generations (ISA extensions), the executed instructions could change, which could silently misalign the BBV and the traced segments.

Do you think we still need these new modes under "manual" tracing mode?

5surim:
Good point about cross-node ISA differences. DynamoRIO traces at the instruction level, so if fingerprinting runs on a node with AVX-512 and tracing runs on a node without it, the compiler-generated code paths could differ, misaligning the BBV segments. In practice, this is mitigated by:

1. **Docker container**: The workload binary is built once inside the image (fixed compiler flags), so the same binary runs on every node. ISA-dependent code paths only matter if the binary uses runtime CPU dispatch (e.g., glibc's `ifunc` resolvers), which is uncommon for most benchmarks.
2. **Slurm constraints**: You can pin all jobs to the same partition/node type with `slurm_options` (e.g., `"--constraint=icelake"` or `"--nodelist=ohm"`).

For this PR specifically, the per-segment fan-out is designed for Slurm (`parallel_segments` mode), but `cluster_then_trace` in manual mode also benefits from the resume support and batched tracing (`TRACE_PARALLEL`). The fan-out itself is Slurm-only — in manual mode, all segments still run sequentially within a single container.

moonseoppkim:
@5surim 

Agreed on (1), but (2) puts the burden on the user, and the failure mode is silent (misaligned BBV/trace, no error). 
Could you at least document it in the codebase or README, so users are aware of this when tracing over Slurm?

5surim:
Good call. Added a warning in README.trace.md under the "Parallel Segment Tracing" section (commit ba1dc08). It explains the cross-node ISA mismatch risk from glibc ifunc resolvers and recommends pinning jobs via slurm_options (e.g., "--constraint=icelake").

moonseoppkim (dismissed):
LGTM

moonseoppkim (approved):
LGTM

PR #401, opened 2026-06-11
State accounting

FlyingFish800:
Implemented 'state files' which store the state (and only the state) of a running process. This allows us to use one authoritative source of information for both the status command, and to check which simpoints to re-run. It also allows the simplification and standardization of these checks. Detailed error reporting is moved to explicit checks after state is determined and static (already failed or succeeded).

Statefiles are stored in {experiment_dir}/logs/statefiles, and contain *only* state information (In progress, Running, Completed, Failed - Scarab, Failed - Runner). Lack of state file means an experiment has not been run.

Runner scrips  are moved to {experiment_dir}/logs/launch_scripts instead of being deleted

### State updates:
- Scarab-infra sci creates the statefile and adds the In progress marker immediately after sbatching
- The launch script immediately writes that it is in the running state as the first thing it does after the slurm/manual scheduler runs it
-  If the docker container is sent exit, terminate, interrupt, or HUP signal, set state to FAILED - Runner
- After scarab completes, the launch script checks the status:
    - If an OOM kill was detected, set state to FAILED - Runner. This has to be handled separately (sigkill is non-maskable)
    - If Scarab returned non-zero exit code or failed to produce csv stats, set state to FAILED - Scarab
    - Otherwise, mark state as successfully completed

Statefiles for tracing is not currently implemented

### DIFF
--- a/scripts/local_runner.py
+++ b/scripts/local_runner.py
@@ -69,9 +69,7 @@ def print_status(user, job_name, docker_prefix_list, descriptor_data=None, workl
         docker_running,
         [],
         dbg_lvl=dbg_lvl,
-        log_file_count_buffer=0,
         strict_log_count=False,
-        log_count_offset=0,
         prep_failed_label="Failed - Prep",
     )
 
@@ -214,7 +212,7 @@ def run_simulation(user, descriptor_data, workloads_data, infra_dir, descriptor_
                         continue
 
                     workload_home = f"{suite}/{subsuite}/{workload}"
-                    write_docker_command_to_file(user, local_uid, local_gid, workload, workload_home, experiment_name,
+                    write_docker_command_to_file(user, local_uid, local_gid, suite, subsuite, workload, experiment_name,
                                                  docker_prefix, docker_container_name, traces_dir,
                                                  docker_home, githash, config_key, config, sim_mode, scarab_binary,
                                                  seg_size, architecture, cluster_id, warmup, trace_warmup, trace_type,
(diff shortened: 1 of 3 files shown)


hlitz:
@yzhao321 Can you take charge of testing this PR before Alex leaves? Try to break it by adding and removing experiments to json during runs, have experiments that fail, reduce memory allocations so that experiments run out of memory, reboot a machine and run slurm jobs on it without executing ./sci --init before, and whatever else you can think of. I'd like to converge to a very stable environment that always works.

hlitz:
@yql5510718 @estodi @yzhao321 did you run some tests with this scarab-infra to test its stability? please let us know your experience

yzhao321:
When I tried this branch to run sim, when I use --status, there are some errors:
image so I cannot track if an job is added during running. I am also wondering the cmd for adding job during running. It is just used '--sim json'? the name of json should be the same as the orig json? Is there any new cmd added for new features? Thanks

hlitz:
@FlyingFish800 can you comment on this? I thought the Log file doesn't match number of accounted jobs should no longer triggered?

hlitz:
> When I tried this branch to run sim, when I use --status, there are some errors: image so I cannot track if an job is added during running. I am also wondering the cmd for adding job during running. It is just used '--sim json'? the name of json should be the same as the orig json? Is there any new cmd added for new features? Thanks

@yzhao321 did this error only happen after adding new config/jobs to the json? Yes you should be able to add a new config to json and rerun ./sci --sim

PR #407, opened 2026-07-19
Fixed `--perf`/`--perf-interactive`

estodi:
Fixed a few issues:
- `--perf_interactive` was not recognized.
- `user_entrypoint.sh` was not sourced in `--perf`.
- `--perf` wrote back the collected data all at once at the end, but when an error occurred in the middle, all data gathered so far was lost.

### DIFF
--- a/sci
+++ b/sci
@@ -4419,6 +4419,7 @@ def main() -> int:
         args.trace,
         args.sim,
         args.perf,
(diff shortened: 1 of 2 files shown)


5surim (approved):
LGTM!

estodi:
Thank you!

PR #409, opened 2026-07-21
Add SPEC 2026 container

estodi:
This PR adds Dockerfile and root/user entry point shell scripts for SPEC 2026. Ubuntu and gcc versions were chosen based on SPEC's original configuration.
~`common/Dockerfile.common` is also modified to correctly handle clang-tidy-18 installation, as ubuntu 22.04 does not have it natively.~ Created sperate PR #433.


hlitz:
hard-to-impossible to review without any description. Especially if so many separate things are packed in the same PR. You need to make the reviewer's life easier. (you can see #411 as an good example)

sci
5surim, line 4422: It has been added in main. It seems the branch is outdated. Please rebase and resolve conflicts.

estodi:
Sorry, this PR included commits from another PR. I fixed the conflicts.

5surim:
Could we make this image `mount-only`, matching what spec2017 just moved to in #411 ?
I'd drop the ADD spec2026.tar.gz + runcpu --action=build from this PR and instead just create the mount point, relying on the descriptor's application_dir to bind-mount a precompiled cpu2026/ at runtime.

We can have a separate dockerfile for anyone with the license but without the precompiled one.

estodi:
Thank you for pointing out. Changed the image to mount-only along with a fix for gcc mismatch.

scripts/utilities.py
```diff
@@ -34,6 +34,15 @@
 BASE_MEMORY_BY_MODE_KEY = "base_memory_mb_by_mode"
 VALID_SIMULATION_MODES = ("memtrace", "pt", "exec")
 
+# Workload images whose system-default compiler is too new for scarab and its
```
hlitz: this comment says that gcc-11 is needed for scdrab but it seems to be needed for spec2026. please clarify. what exactly requires gcc-11?

common/Dockerfile.common
```diff
@@ -25,10 +25,28 @@ RUN apt update && DEBIAN_FRONTEND=noninteractive apt install -y \
     gdb \
     graphviz \
     curl \
-    time \
-    clang-tidy-18 \
- && update-alternatives --install /usr/bin/clang-tidy clang-tidy \
-      /usr/bin/clang-tidy-18 100
+    time
+
+# clang-tidy-18 isn't in every base OS's own repos this file is INCLUDE+'d into
```
hlitz, line 30: seems to be independent of spec26. in principle this should go into a separate earlier PR

scripts/utilities.py
```diff
@@ -424,6 +433,24 @@ def build_scarab_binary(user, scarab_path, scarab_build, docker_home, docker_pre
                 check=True, capture_output=True, text=True)
 
         info(f"Building scarab with image {image_tag_for(docker_prefix, infra_dir)}...", dbg_lvl)
+
+        # Pin an older compiler for images whose default toolchain is too new for scarab.
+        make_env = ""
+        reconfigure = ""
+        cc_cxx = SCARAB_BUILD_COMPILER.get(docker_prefix)
```
hlitz: again, I don't understand how this is related to spec26.
Is this a scarab issue being addressed here? shouldn't we fix scarab so it compiles with new gcc?

hlitz (dismissed):

scripts/utilities.py
estodi: Scarab needs pin 3.1.5, which requires older gcc versions than gcc-15. However, ubuntu jammy's default gcc (gcc-11) does not cause any trouble and this fix was needed only because I made gcc map to gcc-15 in spec 2026 image, which is not strictly needed so long as we specify the compiler version in the compile option. Reverted to the original.

scripts/utilities.py
estodi: gcc-11 is used for scarab build while SPEC 2026 uses gcc-15.

5surim (approved):
LGTM

5surim (approved):
LGTM!

common/Dockerfile.common
estodi: Addressed in 45acd6a

PR #412, opened 2026-07-21
ci: build and publish the spec2017 image to ghcr (like allbench_traces)

hlitz:
## What

Extends the Docker image CI (`.github/workflows/docker-image.yml`) to build and publish the **spec2017** workload image to `ghcr.io/litz-lab/scarab-infra/spec2017:<githash>`, mirroring how `allbench_traces` is published. Also adds `workloads/spec2017/` to the change-detection filter so a spec2017 (or `common/`) change triggers a rebuild.

## Why

Today only `allbench_traces` is published, so `sci` can pull it and skip building. For `spec2017` there's nothing to pull, so `sci` falls back to building the image **locally** on each machine/compute node. A local build uses the repo dir as the docker build context, which includes the ~35 GB `scarab_builds/` artifact cache — bloating the build and OOM-killing per-node builds under the Slurm job memory limit.

Publishing spec2017 means nodes **pull** the prebuilt image (built in CI from a clean checkout, tiny context) and never build locally — same model as `allbench_traces`.

This became possible only after the spec2017 image stopped baking the licensed SPEC tarball (see dependency below); the published image contains no licensed content.

## Dependency / rollout

- **Depends on #411** (remove the `spec2017.tar.gz` bake). CI builds from a clean checkout with no licensed tarball, so the spec2017 image only builds once #411's Dockerfile is on `main`. **Merge #411 first.**
- After the first successful push, make the `ghcr.io/litz-lab/scarab-infra/spec2017` package **public** (org → Packages) so nodes can pull without auth, as with `allbench_traces`.

## Notes

- Both images are built/pushed under the same `changed` condition and share `last_built_tag.txt`, so their tags stay aligned at the same commit hash.
- No `sci` change needed — its pull path is already generic over workload group.

### DIFF
--- a/.github/workflows/docker-image.yml
+++ b/.github/workflows/docker-image.yml
@@ -46,7 +46,7 @@ jobs:
             echo "changed=true" >> $GITHUB_OUTPUT
           else
             git fetch origin main --depth=50
-            CHANGED=$(git diff --name-only "$LAST_TAG" HEAD | grep -E '^(common/|workloads/allbench_traces/)' || true)
+            CHANGED=$(git diff --name-only "$LAST_TAG" HEAD | grep -E '^(common/|workloads/allbench_traces/|workloads/spec2017/)' || true)
             if [[ -n "$CHANGED" ]]; then
               echo "Relevant changes detected:"
               echo "$CHANGED"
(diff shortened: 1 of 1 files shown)


.github/workflows/docker-image.yml
```diff
@@ -46,7 +46,7 @@ jobs:
             echo "changed=true" >> $GITHUB_OUTPUT
           else
             git fetch origin main --depth=50
-            CHANGED=$(git diff --name-only "$LAST_TAG" HEAD | grep -E '^(common/|workloads/allbench_traces/)' || true)
+            CHANGED=$(git diff --name-only "$LAST_TAG" HEAD | grep -E '^(common/|workloads/allbench_traces/|workloads/spec2017/)' || true)
```
5surim: Should we build a new image for allbench_traces or spec2017 if one of them changes?
I think we don't need to build spec2017 if only allbench_traces changes.
But still, in most cases, there are changes in common, so it will anyway build new images for allbench_traces and spec2017.

5surim (approved):
LGTM

PR #417, opened 2026-07-22
spec2017 exec: private local run dir + drop shrc (read-only app tree)

hlitz:
## Problem
Exec-driven SPEC2017 sims fail on every node with:

```
You are not allowed to write into /tmp_home/application/cpu2017/config.
```

The SPEC binaries are now bind-mounted **read-only** from NFS (see #411), and `/soe/hlitz` NFS is full. Two independent writes into that shared tree were the cause:

1. **`source shrc`** in `workloads/spec2017/workload_user_entrypoint.sh` — SPEC's `runcpu` bootstrap, which insists `$SPEC` be writable. We never run `runcpu`; the launch command (`workloads_db` `binary_cmd`) invokes the benchmark by **absolute path** with **no env from shrc** (`env_vars: null`), so it's pure ceremony.
2. **The benchmark's own file writes** land in the shared run dir. `run_exec_single_simpoint.sh` `cd`'d into the benchmark's run dir, and — critically — SPEC `binary_cmd`s embed **absolute paths for the binary, inputs AND outputs** (e.g. `gcc_r ... -o $tmpdir/.../foo.s`), so the output is written by absolute path into the shared tree regardless of cwd. Even when the tree was writable, concurrent simpoints/users clobbered each other.

## Fix
1. Drop `source shrc` (entrypoint becomes a documented no-op).
2. Run each simpoint in a **private, node-local** run dir:
   - make a real local **copy** of the reference run dir (`cp -rL --preserve=timestamps` + `chmod -R u+rwX`),
   - rewrite the command's reference-dir prefix (with `$tmpdir` expanded) to the private dir, so the binary/inputs/outputs all resolve locally,
   - `cd` there and clean up on exit.

Local base is overridable via `SCARAB_RUN_LOCAL_TMP` (default `/tmp`, i.e. the container overlay — which is node-local disk).

### Why a real copy, not symlinks
A symlink-the-inputs approach was tried first and fails in practice: the shared run dirs are **polluted with prior-run output files** (arbitrary benchmark-specific names). A symlink to such a file points at a read-only NFS target, and the benchmark overwriting it fails with EACCES. The clean copy-on-write primitive (overlayfs / bind CoW) needs root, but the benchmark runs **unprivileged**, so we use a real copy. Run dirs are small (tens to ~200 MB) and removed on exit.

## Verification
- Old `not allowed to write` / EACCES failures: **gone**.
- Baseline SPEC2017 exec simpoints complete with full `core.stat.0.out`.
- (Separately observed: the load-prediction scarab configs hit an unrelated assertion in `map_rename.c` — a scarab-code issue, not part of this infra PR.)

## Scope / risk
`run_exec_single_simpoint.sh` is shared by all exec-driven sims; the private-copy change is strictly safer for all of them (isolation + no shared-tree writes). Completes the read-only-mounted-app work started in #411.

### DIFF
--- a/common/scripts/run_exec_single_simpoint.sh
+++ b/common/scripts/run_exec_single_simpoint.sh
@@ -44,8 +44,23 @@ PIN_EXEC="$SCARABHOME/src/pin/pin_exec/obj-intel64/pin_exec_${SCARAB_BIN}.so"
 cp "$PIN_EXEC" "$OUTDIR/$segID/pin_exec.so"
 
 cd $OUTDIR/$segID
-BINARY_DIR=$(dirname ${BINCMD%% *})
-cd $BINARY_DIR
+
+# Run the benchmark with cwd = a PRIVATE, node-local run dir. This keeps the
+# benchmark's output-file writes off the shared app tree, which is bind-mounted
+# read-only from NFS: writing there fails (and, when it didn't, concurrent
+# simpoints/users clobbered each other's outputs). Inputs are symlinked in
+# (shared read-only, no copy); the binary is launched by absolute path from
+# BINCMD, so cwd only governs where the benchmark's own output files land.
+REFDIR=$(dirname ${BINCMD%% *})
+RUNDIR="${SCARAB_RUN_LOCAL_TMP:-/tmp}/scarab_run_${SCARAB_BIN}_$$_${segID}"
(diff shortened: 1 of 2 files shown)


moonseoppkim (approved):
LGTM

PR #410, opened 2026-07-21
Upgrade `--perf`

estodi:
This PR extends perf collection to gather raw hardware counters alongside the existing top-down (toplev) and elapsed-time measurements, and add a counter-multiplexing pre-flight check.

This PR is independent, but built on top of #409.

New metrics
- Add HW counters: instructions, cycles, branches, branch-misses,
  L1-icache/L1-dcache/L2 (l2_rqsts.miss)/LLC-load-misses, iTLB/dTLB-load-misses.
- Collect them in the SAME `perf stat` pass as elapsed time (added `-e`), so no extra workload run is needed.
- New parse_perf_counters(): parse raw event counts (handles thousands separators, skips <not counted>/<not supported>).
- New compute_hw_metrics(): derive ipc, branch_pct, branch_mispred_rate, branch_mpki, l1i/l1d/l2/l3_mpki, itlb/dtlb_mpmi, per-run `instructions`, and min_coverage_pct. Missing source events yield None.
- run_perf_stat_for_config() now returns (elapsed, hw_metrics).

Counter-budget check (multiplexing guard)
- New check_counter_budget() run once before collection: probes whether the event set fits without multiplexing via _probe_counter_fit() on a trivial workload; parses coverage with new parse_perf_min_coverage().
- If it multiplexes, diagnose the cause (SMT via _read_smt_active(), NMI watchdog via _read_nmi_watchdog()) and prompt to continue or abort (_prompt_continue_or_abort(); aborts on non-interactive stdin).

Descriptor-configurable knobs
- perf_core, max_repeats, target_seconds in the descriptor json now override PERF_CORE / PERF_REPEAT_DEFAULT / PERF_TARGET_SECONDS.

Output schema (workloads_db.json performance block)
- Add `hw_metrics` and `repeat_count`; log hw_metrics during collection.

PR #422, opened 2026-07-25
Add SPEC 2026 perf results

estodi:
This PR updates `workloads_db.json` with SPEC 2026 perf and a descriptor json to reproduce it, using #410 on #409.


5surim (approved):
Looks really good!

workloads/spec2026/spec2026_perf.json
```diff
@@ -0,0 +1,197 @@
...
+  "image_name":"spec2026",
+  "perf_core":7,
+  "max_repeats":2,
+  "target_seconds":3600,
+  "perf_configurations":[
+    {
+      "workload":"stockfish_r",
+      "suite":"spec2026",
+      "subsuite":"rate_int",
+      "binary_cmd":"runcpu --config=memtrace.cfg --action=onlyrun 706.stockfish_r",
```
5surim, line 18: Hi @estodi ,

I missed this. I think the binary_cmd should be an absolute path.
Can you check spec2017 and 2006 for it?
For the exec-driven simulation, the command is attached to scarab command where the `runcpu` is not always there.

Thanks,
Surim

estodi: Addressed in PR 438

PR #423, opened 2026-07-27
Mount infra scripts and tag images by content, so only real image changes rebuild

hlitz:
Two related changes so that a scarab-infra change only costs an image rebuild when the image actually differs.

## 1. Mount the workflow scripts instead of baking them in

`common/Dockerfile.common` COPY'd 12 scripts into the image at line 49 — *above* the DynamoRIO, pin and SimPoint builds — so editing a one-line shell script invalidated that layer and everything under it, forcing a full `--no-cache` rebuild of a multi-GB image. Every call site then worked around the resulting staleness by `docker cp`-ing the host copies back over the baked ones, with eight separately maintained copies of the file list.

Those lists had already drifted: `run_simulation.py` copied `run_memtrace_single_simpoint.sh` to `/usr/local_bin` (typo), so memtrace interactive shells silently ran the stale baked script.

Now the host checkout is bind-mounted read-only at `/scarab_infra` on every `docker run`, and `root_entrypoint.sh` — invoked through its mounted path — symlinks the scripts into `/usr/local/bin`. It globs rather than listing filenames, so a new script needs no bookkeeping. Keeping `/usr/local/bin` as the published location means the ~15 absolute references elsewhere (`run_clustering.sh`, the `source` line appended to `~/.bashrc`, …) are untouched, and the directory stays writable for `perf_entrypoint.sh`'s `perf` symlink.

Images now own only the toolchain and the workload sources.

**No new deployment requirement under Slurm.** The generated sbatch scripts already embedded `infra_dir` literally and ran `docker cp` from it on the compute nodes, so the checkout already had to be at the same absolute path everywhere (NFS).

## 2. Tag images by content, not by git hash

Mounting alone would not have removed the rebuild. The tag was the scarab-infra git hash, which is not the image's identity: every infra commit minted a new tag for byte-identical content, so `prepare_docker_image()` saw a missing tag and called `--build-image`, which then had to materialise the tag on each node — retag a local base image, or, if that base was absent (the normal case on a fresh Slurm node), pull multiple GB from ghcr purely to rename it. It also accumulated a tag per commit; five identical `spec2017` images under five hashes was typical.

The tag is now a hash of exactly the repository content that `docker build` bakes in — the Dockerfiles, `fingerprint_src/`, and the workload directory, excluding the scripts that are bind-mounted at run time. Identical content yields an identical tag on every node and every branch, so `image_exist()` simply hits and nothing else runs.

That collapses the machinery built around the wrong key:

- `last_built_tag.txt`, the diff against it, and the retag path are gone. `--build-image` is now "does this tag exist locally? else pull it, else build it". The CI auto tag-bump PR that maintained the file goes with it.
- The dirty-worktree guard is gone. It existed because a dirty build would have been mislabelled with the committed git hash; hashing the working tree makes that impossible, so a local Dockerfile edit builds under its own tag and you can iterate without committing.
- The generated sbatch scripts no longer pass `--githash`; the node derives the same hash from the checkout it already `cd`s into.

## Bugs fixed along the way

- **`fingerprint_src/` never triggered a rebuild.** `Dockerfile.common` COPYs it in and compiles it into `libfpg.so`, but neither `sci`'s rebuild diff nor the CI grep covered it, so edits to the fingerprint client silently reused a stale `libfpg.so`. It is now part of the image identity.
- **`/usr/local_bin` typo** in `run_simulation.py` (described above) — disappears with the refactor.
- **`Dockerfile.oss` / `Dockerfile.tailbench` were unbuildable**, COPYing `run_scarab.sh`, `run_exp_using_descriptor.py`, `gather_cluster_results.py` and others that no longer exist under `common/scripts`. Removing the COPY blocks fixes that as a side effect.

Also removed `write_docker_command_to_file_run_by_root()` (no caller anywhere, and it never invoked `root_entrypoint.sh`, so it could not have worked under the mount) and `scripts/update_scripts.sh` (the manual version of the same `docker cp` dance).

## Verification

Against a real container (`spec2017:ff6c6d8`):
- the symlinks replace the baked scripts, including the per-workload `workload_*_entrypoint.sh`
- a host edit is visible in a running container with no rebuild and no `docker cp`
- the mount is read-only, and `/usr/local/bin` stays writable
- `user_entrypoint.sh` sources with the correct env
- a container started without the mount fails with an actionable message

Tag behaviour:

| change | spec2017 | allbench_traces |
|---|---|---|
| any bind-mounted script | unchanged | unchanged |
| `workloads/spec2017/Dockerfile` | rebuilds | unchanged |
| `common/Dockerfile.common` | rebuilds | rebuilds |
| `fingerprint_src/` | rebuilds | rebuilds |
| uncommitted Dockerfile edit | own tag; reverting restores the original | — |

The host, the generated sbatch script, and the node-side `prepare_docker_image` all agree on the tag.

## Breaking change

Existing `<workload>:<githash>` images do not match the new tags, so the first run after this lands rebuilds or pulls once per workload, and CI republishes under content tags.

## Caveat and follow-ups

The hash covers build *inputs*, not resulting image bits. Several fetches are unpinned — the DynamoRIO tag, the SimPoint and pmu-tools clones, `FROM ubuntu:focal`, the `litz-lab/scarab` clone that feeds `pip3 install -r requirements.txt`, and `apt` — so two builds of the same tag can still differ. That caveat is unchanged from the git-hash tag, and it only bites when a rebuild actually happens.

Deliberately left for separate PRs:

- **Pinning** those fetches, so a rebuild reproduces the same image. `apt` is the residual that cannot practically be closed.
- **Digest provenance**: nothing currently records which image an experiment ran (container names carry the *scarab* git hash, not the image), so if bits ever do drift it is not detectable after the fact.

`scripts/image_identity.py` deliberately imports only the standard library, so CI can compute tags from a bare checkout via `python3 -m scripts.image_identity <workload>`.


hlitz:
@5surim I already did a review and some testing but please have another look, if tyou think the general architecture makes sense. This will eliminate almost all image rebuilds.

5surim (commented):
Reviewed against the `mount-infra-scripts` branch. The reasoning behind both halves is sound and the diff is unusually well-commented — the tag genuinely was the wrong key, and `fingerprint_src/` really was invisible to both the old `sci` rebuild diff and the CI grep. One blocker and a short list of correctness issues below.

## Blocker

**`scripts/utilities.py:2605` — `prepare_trace()` crashes immediately.**

```python
rebuild_scarab(infra_dir, scarab_path, user, docker_home, docker_prefix, githash, scarab_githash, scarab_build, stream_build=False, dbg_lvl=dbg_lvl)
```

`githash` was dropped from `prepare_trace`'s signature (line 2553) but not from this call. Two faults at once: `NameError: name 'githash' is not defined`, and 8 positionals into the now 7-parameter `rebuild_scarab` (line 979). This is the non-interactive branch — every normal `./sci --trace` run, local and Slurm — and it fires before any work happens.

```python
rebuild_scarab(infra_dir, scarab_path, user, docker_home, docker_prefix, scarab_githash, scarab_build, stream_build=False, dbg_lvl=dbg_lvl)
```

It's the only stale call site; I checked every `prepare_simulation` / `prepare_trace` / `rebuild_scarab` / `build_scarab_binary` / `lint_scarab_binary` / `write_*_docker_command_to_file` caller on the branch and the rest are consistent. The Verification section covers mounts and tag behaviour thoroughly but evidently never exercised `--trace`.

## Content hash — cases that can silently serve the wrong image

**Untracked files are invisible to the tag but visible to `docker build`.** `_tracked_files()` uses `git ls-files`, so a new, not-yet-`git add`ed file under `workloads/<group>/` is baked into the image without changing the tag. The docstring's claim — "Hashing the working tree rather than the git index means an uncommitted Dockerfile edit produces a different tag, so a local experiment cannot silently reuse an image built from different sources" — holds for *modified tracked* files but not for *new* ones. Two developers with different untracked build-context files get the same tag for different images, and each other's cached image. Either hash untracked-but-not-ignored files too (`git ls-files --others --exclude-standard`), or narrow the docstring to say so explicitly.

**`image_context_pathspec()` is a hand-maintained approximation of the build context, and nothing enforces it.** `workloads/oss/Dockerfile:35` already `COPY`s `./workloads//OSS/setup.sh` — outside `workloads/oss`, therefore outside the hash. `oss` is unbuildable anyway so it's latent, but it's exactly the failure mode that will recur. Since this PR already edits `docs/README.add_workload.md`, one line there would help: *if your Dockerfile `COPY`s anything outside `workloads/<suite>/`, update `image_context_pathspec()`.*

**File mode isn't hashed.** `COPY` preserves the exec bit, so `chmod +x` on a copied file changes the image but not the tag. Git tracks the bit; folding the `ls-files -s` mode into the digest is a two-line fix. This belongs alongside the unpinned-fetch caveat already documented.

**`_HASH_CACHE` is never invalidated.** The inputs are tiny (spec2017: 10 files / 176K; autoware: 46 / 308K), so the cache buys almost nothing while letting a long-lived `sci` process hold a stale tag across an edit. I'd drop it.

## Runtime / mount

**Unset `APP_GROUPNAME` degrades silently.** `root_entrypoint.sh` globs `"$INFRA_HOME/workloads/$APP_GROUPNAME"/workload_*_entrypoint.sh`; if the var is empty the path becomes `workloads//workload_...`, `[ -f ]` fails, and the per-workload entrypoints are simply never published. Every current `docker run` site does set it — and the PR correctly adds `-e APP_GROUPNAME` to `build_scarab_binary` and `lint_scarab_binary` — but given the mount-missing case gets a good actionable error, this one deserves at least a warning rather than a no-op.

**Pre-existing `--perf` containers will silently keep using their baked scripts.** They carry the `/tmp_home/.scarab_perf_ready` sentinel and have no `/scarab_infra` mount, and `sync_perf_support_files()` is gone, so nothing ever refreshes them. Worth adding to the Breaking-change section that these need a `docker rm`.

**Stale symlinks are never pruned.** In a reused container, a renamed or deleted host script leaves a dangling `/usr/local/bin/<old>`. A `find /usr/local/bin -xtype l -delete` before the loop would close it.

**Two sources of truth for the mount target** — `ENV INFRA_HOME="/scarab_infra"` in the Dockerfiles vs `INFRA_MOUNT_TARGET` in `utilities.py`. They can drift silently; changing the Dockerfile ENV wouldn't move the mount. A cross-reference comment at minimum.

## Verdict

Approve once `utilities.py:2605` is fixed. The untracked-files gap is the one design issue I'd want either resolved or explicitly documented before merge; the rest are follow-ups.

hlitz:
Thanks @5surim — all nine findings verified independently and fixed in e930fa8.

**Blocker.** Confirmed and fixed. One correction to the diagnosis: `rebuild_scarab` takes 9 parameters, not 7, so the second fault was `stream_build` receiving both a positional and a keyword value rather than an arity overflow. The `NameError` fires first either way, so the impact and the fix are exactly as you described.

It slipped my verification because the final sweep filtered `githash` hits with `grep -v scarab_githash` and that line contains `scarab_githash`; `py_compile` cannot see a `NameError`; and the runtime check exercised the two sbatch writers but never `prepare_trace`. Both gaps are now closed by two static scans run over every changed file and diffed against `origin/main` — undefined names, and arity/duplicate arguments. I validated them against the pre-fix code: each independently reports line 2605. On the branch both now match `origin/main` exactly (two pre-existing `err()` calls, one pre-existing undefined name). `prepare_trace` is also now exercised at runtime and `rebuild_scarab` binds against the real signature.

**Untracked files — resolved, not documented.** You were right that it is live rather than theoretical: I reproduced it, and `COPY fingerprint_src $tmpdir/fingerprint_src/` is a whole-directory copy, so an un-added file really is baked in. Now hashing tracked plus `--others --exclude-standard`. One wrinkle worth flagging: `--exclude-standard` applies `.gitignore`, which did not carry `.dockerignore`'s junk patterns, so editor droppings started churning the tag. `.gitignore` now carries them, with `#*#` escaped — unescaped, git reads it as a comment.

**Also fixed:** exec bit folded into the digest; `_HASH_CACHE` dropped; `perf_container_initialized()` now requires the mount, so a pre-existing `--perf` container gets the actionable error instead of silently running baked scripts (this is better than the docs note you suggested — no `docker rm` needed); dangling symlink pruning restricted to links pointing into `$INFRA_HOME` so `perf_entrypoint.sh`'s `perf` link survives; `APP_GROUPNAME` unset or unknown now warns; `image_context_pathspec()` documents that it is a hand-maintained approximation and names `workloads/oss/Dockerfile` as the existing violation, with a matching rule in `README.add_workload.md`; the three declarations of the mount path cross-reference each other.

Pinning the unpinned fetches and recording image digests for provenance remain follow-ups.

5surim (approved):
LGTM

PR #430, opened 2026-08-18
Fix --trace parallel segments mode

estodi:
This PR fixes a few issues in  `--trace` parallel segments mode
- Fix slurm_options were not passed down to the later phases. --mem is handled by infra, now adjustable via env_vars, and ignored if present in slurm_options.
- Skip workloads that are already fully traced on submission, rather than submit a job and let it fail.
- Give names to phase2/3 jobs so they clearly tell thier workload and phase in queue.
- Forbid updating workload_db when there are failed workloads so it will not block re-run.
- Remove json read that is never used.
- Add a few status messages


5surim (approved):
LGTM Thanks

PR #431, opened 2026-08-18
Fix hard coded relative inputs

estodi:
Some workloads read inputs using hard-coded relative path, which users cannot / are not supposed to change. Current infra supports mounting the application directory without polluting the original, but the feature is limited to workloads that take inputs through their runtime arguments. This PR fixes the issue by copying the files inside the workload's dir to a local run dir. The application binary itself will not be copied as the location of the binary does not matter while its size can be huge. The new feature is optionally enabled by `SCARAB_COPY_LOCAL_RUNDIR=1` in env_vars; by default, the current behavior is preserved.


hlitz:
@estodi quick questions.

1. if inputs are read-only, do we really need to copy them, ie. as long as we don't write it may be fine multiple processes reading the same data at the same time
2. if the paths are the problem would it be sufficinet to symlink instead of copy?

estodi:
Thank you for the questions
1. Copying input files itself is not strictly required as long as the input files exist in the correct path relative to CWD. It is coming from the fact we cannot run the binary directly in the application dir. yes, as long as we don't write it may be fine multiple processes reading the same data at the same time.
2. Fixed in the newer commits

hlitz (approved):
LGTM

PR #432, opened 2026-08-18
Fix --trace drrun overshoot and broken minimize resume

estodi:
drrun records roi_length instructions (from roi_start), but this limit specified by `-trace_for_instrs` is rough and it often overruns. Sometimes it goes too far that drraw2trace cannot finish processing the produced raw file, leaving a broken zip without the central directory. As a result, minimize stage kept failing on some large workloads. This PR fixes the issue by setting a hard limit on how much drrun executes via `-exit_after_tracing`. This fix also reduces the runtime of drrun because it terminates after tracing the requested window, whereas it previously continued to run until the end even after stop recording.

This PR also addresses an issue where the minimize stage was running on segments that were already minimized.


hlitz:
@moonseoppkim what was the result of our discussion here? Is it fine to exit after tracing if we perform the tracing in a temp dir and so it does not matter that the app did not finish?

moonseoppkim:
@hlitz
My concern was that using the same directory for multiple runs could cause them to interfere with each other
(artifacts generated by a previous run might affect the next iteration.)

it should be fine as long as we give each run its own directory after copying.

@estodi 
However, I have a question about the"exit_after_tracing" change you added to terminate the process. 
In this repo (https://github.com/5surim/dynamorio/commits/release_10.0.0/), 
more ref. https://github.com/DynamoRIO/dynamorio/pull/7772

I made a change so that fetched instructions are counted correclty at scarab with trace_after_instrs and trace_for_instrs. 
I'm not confident that "exit_after_tracing" counts fetched instructions the same way trace_for_instrs does.

If you want to add this change, 
please add more validations in terms of "exit_after_tracing" can handle "REP" instruction correctly, like trace_for_instrs.

hlitz:
@estodi what is the status of this PR? please respond to @moonseoppkim

estodi:
Thank you for your comments.

I dug into the issue and found the root cause in dynamorio, more precisely, in drcachesim. However, this PR is independent of that bug and works correctly even in the presence of the dynamorio’s bug. It is ready for review.

The issue that causes drcachesim to overrun the requested tracing window is its incorrect instruction count. As part of the fix for REP counting, drcachesim introduced REP detection and changed the way it counts instructions. To determine if it has covered the requested window, it accumulates the number of instructions read so far, as shown in the code below. However, instead of adding the number of instructions inside a basic block, it always adds only one. As a result, it consistently undercounts instructions by roughly the number of instructions in a basic block and traces far beyond the requested window.
https://github.com/5surim/dynamorio/blob/release_10.0.0/clients/drcachesim/tracer/output.cpp#L1165-L1178

The good news is that this bug in drcachesim affects only the mechanism for limiting the tracing window and does not affect the quality of the raw output. Although drcachesim overruns the requested window quite dramatically, this merely produces a much larger raw output covering a longer instruction window, without compromising its quality.
Because drraw2trace only reads the raw output, the correctness and consistency of its output are also unaffected. As long as the region of interest is covered, that is sufficient in our case. 
Whether the ROI is covered can be determined by checking the number of chunks produced by drraw2trace, as the number of instructions per chunk is specified. Therefore, this PR adds a safeguard at the beginning of the minimize stage, right after drraw2trace, to guarantee that the roi is covered. This way we can trust the trace file, regardless of whether drcachesim counts instructions correctly.

As you pointed out, exit_after_tracing counts references rather than instructions. However, this does not pose a problem as long as execution extends beyond the ROI. Execution beyond the ROI is realized by TRACE_ROI_HEADROOM (3x by default), and the added safeguard verifies that the ROI is fully covered.

moonseoppkim:
@estodi Thanks for your in-depth analysis.

One question on the `TRACE_ROI_HEADROOM` part: what does the 3x multiply, 
and where does this number come from?

From #264, segment 518 of 557.xz_r (test input) needs 77.4M instructions to cover a 10M-fetched window:

```
trace_then_cluster log
/soe/mkim293/seop/simpoint_flow/557.xz_r/xz_r_baseline/fingerprint/518/sim.log
77391485 (number of total instructions)
10000000 (number of fetched instructions)

/soe/mkim293/seop/temp/trace_then_cluster_full/557.xz_r_full/cluster_518/518/inst.stat.0.out
Periodic:          Cycles: 71674686              Instructions: 77391483              IPC: 1.07976

ST_NOT_CF                                    217502426       75.969%        227843345       76.765%
ST_CF_BR                                        182652        0.064%           187629        0.063%
ST_CF_CBR                                      1121756        0.392%          1273185        0.429%
ST_CF_REP                                     67391593       23.538%         67391593       22.706%    <<<<<<<<<<
ST_CF_CALL                                       22698        0.008%            25190        0.008%
ST_CF_IBR                                        47305        0.017%            47305        0.016%
ST_CF_ICALL                                       5690        0.002%             5690        0.002%
ST_CF_ICO                                            0        0.000%                0        0.000%
ST_CF_RET                                        29656        0.010%            32139        0.011%
ST_CF_SYS                                           15        0.000%               15        0.000%
                                             286303791      100.000%        296806091      100.000%
                                                  0.72         1.12              0.69         1.11

557.xz_r seg 518, trace_then_cluster ground truth:
  fetched instructions : 10,000,000
  total instructions   : 77,391,485    -> 7.74x
```

There is at least 7.7x and likely above 10x. 

If the budget is `3 * roi_length`, drrun would exit before covering the requested ROI (warmup_chunks + 1 chunks),
and check_traced_roi would fail.  So 3x looks like a heuristic rather than a lower bound, and the failure mode is that this segment becomes unobtainable.

+ As you said, this change is not affecting tracing quality. 
It's just under-counting the fetched instruction counts for the roi_window.

**On the proper fix suggestion**

I'd rather fix the window counter in `output.cpp` than work around it(trace_after_instr), 
and align `-exit_after_tracing` with the same counting.

+ backing experiment numbers are strongly required to verify your changes.

hlitz:
I still have another question. Are you saying: Whenever drcachesim detects
a REP, it will count only 1 instruction, independently of the BBL's size?
That would be  a weird bug. can we fix it?

moonseoppkim:
@hlitz 

A REP always forms a size-1 basic block, so counting 1 is correct.
(* there is no case where a REP shares a 3-4 instruction block with other instructions. So this part is not a bug.)

hlitz:
then I do not understand @estodi s message. It does not seem that there is
a bug. What is the reason then for the miscounting? Why can;t we just count
the reight number of instructions instead of overshooting?

is the only problem that exit_after_tracing counts references not inst? or
is there something else? I then don't know how 3x is safe ie (if the ratio
of REPS is greater 33%)?

estodi:
The problem is not limited to REP. drcachesim only counts the number of basic blocks. If a basic block happens to go back to the same basic block start address, it will also be ignored.
`addr_t pc = instru->get_entry_addr(drcontext, mem_ref);` gives the start address of the basic block `mem_ref` records. So, it accumulates only one per basic block, and whether it is REP or just a usual basic block does not matter here.
```
static thread_local addr_t prev_pc = 0;
for (mem_ref = data->buf_base + header_size; mem_ref < buf_ptr;
     mem_ref += instru->sizeof_entry()) {

    uintptr_t toadd = 0;
    trace_type_t type = instru->get_entry_type(mem_ref);

    if (type == TRACE_TYPE_INSTR) {
        addr_t pc = instru->get_entry_addr(drcontext, mem_ref);

        if (pc != prev_pc) {
            toadd = 1;
            prev_pc = pc;
        }
    }

    hit_window_end = count_traced_instrs(
        drcontext, toadd, op_trace_for_instrs.get_value());
```
Instead, we can have something like this:
```
        if (pc != prev_pc || instru->get_instr_count(mem_ref) > 1) {
            toadd = instru->get_instr_count(mem_ref);
            prev_pc = pc;
        }
```
(The same REP recorded conesecutively after the first time satisfies `pc == prev_pc && instru->get_instr_count(mem_ref) == 1`.)

moonseoppkim:
@estodi

Good point, not REP-specific. Any self-looping block hits it?

A `for` or `while` body ends in `jne` back to its own start, 
so every iteration records a PC entry at the same address and only the first one counts.
Is that right?

(So with `op_count_fetched_instrs` we undercount the tracing window, the window closes late,  and the raw output ends up much larger than requested?? Not a trace quality issue, just oversized raw output?)

Could you back this with experiment numbers? A sample case would help, 
showing how many instructions are missed in the tracing window under`op_count_fetched_instrs`.

I think dynamorio is quite complicate, so we need to verify your idea with sample code before submitting changes.
(I already had permission from surim for the custom DynamoRIO repo, so feel free to open a PR there.)

estodi:
> A for or while body ends in jne back to its own start,
so every iteration records a PC entry at the same address and only the first one counts.
Is that right?

Yes, if the loop does not jump to somewhere else, that's what happens.

> (So with op_count_fetched_instrs we undercount the tracing window, the window closes late, and the raw output ends up much larger than requested?? Not a trace quality issue, just oversized raw output?)

Correct, drcachesim assumes the window is not hit yet even when it already does and continues to produce the raw output past the window end.

Let me get back to you with additional experiments.

estodi:
I counted the running total of every `toadd` in both the current and fixed versions of drcachesim using the following settings:

| Setting | theoretical instruction counts | current drcachesim | fixed drcachesim |
| ---|---|---|---|
| 100 nops, no control flow | 104 | 1 | 104 |
| single-block loop, back-edge to own head | 10004 | 3 | 10004 |
| multi-block loop, body split by a never-taken jz | 11004 | 2001 | 11004 |
| rep stosb(64) in a 1000-trip loop | 7004 | 3001 | 7004 |

## 100 nops, no control flow

This is a minimal example without a loop. It contains 104 instructions, but the current drcachesim counts only one instruction.

```
_start:
    xor     %eax, %eax          /* 1 */
    .rept 100
    nop                         /* 100 */
    .endr
    mov     $60, %eax           /* 1  SYS_exit */
    xor     %edi, %edi          /* 1 */
    syscall                     /* 1 */
```

## single-block loop, back-edge to own head

This is a case that the current version of drcachesim accidentally ignores. The program contains a loop that jumps back to its own entry point. As a result, the loop satisfies `pred_pc == pc` and is skipped by the current drcachesim. Consequently, it counts only three instructions, whereas the correct count is `1 + 1000 * (8 + 1 + 1) + 3 = 10004`. The fixed version detects the loop using the condition `instru->get_instr_count(mem_ref) > 1` and counts the instructions correctly.

```
_start:
    mov     $1000, %ecx         /* 1 */
L:
    .rept 8
    nop                         /* 8x 1000 */
    .endr
    dec     %ecx                /* 1000 */
    jnz     L                   /* 1000 */
    mov     $60, %eax           /* 1 */
    xor     %edi, %edi          /* 1 */
    syscall                     /* 1 */
```

## multi-block loop, body split by a never-taken jz

This case satisfies pred_pc != pc. Therefore, the current version of drcachesim correctly counts the number of basic blocks as 2001, but incorrectly uses this value as the instruction count. The actual number of instructions is 11004.

```
_start:
    mov     $1000, %ecx         /* 1 */
L:
    .rept 4
    nop                         /* 4x 1000 */
    .endr
    test    %ecx, %ecx          /* 1000 */
    jz      done                /* 1000  never taken (ecx != 0 here) */
M:
    .rept 3
    nop                         /* 3x 1000 */
    .endr
    dec     %ecx                /* 1000 */
    jnz     L                   /* 1000 */
done:
    mov     $60, %eax           /* 1 */
    xor     %edi, %edi          /* 1 */
    syscall                     /* 1 */
```

## rep stosb(64) in a 1000-trip loop

This program shows that the current drcachesim handles REP correctly, although it still miscounts the number of instructions within a basic block. The correct instruction count is `1 + 1000 * 7 + 3 = 7004`, but drcachesim reports 3001 because it counts the following groups: `mov; lea; mov; xor; cld`, `rep`, `dec; jnz`, 999x (`lea; mov; xor; cld`, `rep`, `dec; jnz`), `mov; xor; syscall`.

```
_start:
    mov     $1000, %r8d         /* 1 */
L:
    lea     buf(%rip), %rdi     /* 1000 */
    mov     $64, %ecx           /* 1000 */
    xor     %eax, %eax          /* 1000 */
    cld                         /* 1000 */
    rep     stosb               /* 1000, not 64000 */
    dec     %r8d                /* 1000 */
    jnz     L                   /* 1000 */
    mov     $60, %eax           /* 1 */
    xor     %edi, %edi          /* 1 */
    syscall                     /* 1 */
```

moonseoppkim:
@estodi 

Thanks, Sun!
Looks good to me, Could you open a new PR at https://github.com/5surim/dynamorio/tree/release_10.0.0?

hlitz:
@moonseoppkim could you LGTM this PR if you think its ok

moonseoppkim:
@estodi 

I think the current PR's headroom isn't a good idea,

1. We'd need a new PR upstream in DynamoRIO for how the tracing window is counted (trace_after_instr).
2. We'd also need a matching change in exit_after_tracing, similar to what we did for trace_after_instr.
(If you really want to replace the parameter from "trace_after_instr" to "exit_after_tracing"

hlitz:
I agree with @moonseoppkim assesment. DynamoRIO is broken.

PR #438, opened 2026-08-25
Add correct SPEC 2026 perf result

estodi:
This PR corrects SPEC 2026 perf result using absolutized binary commands instead of the runcpu wrapper. When runcpu has multiple invocations with different configurations, each invocation is recorded as a separate workload with a numeric suffix, as in the other suites.

### DIFF
--- a/workloads/spec2026/spec2026_perf.json
+++ b/workloads/spec2026/spec2026_perf.json
@@ -4,7 +4,9 @@
   "_user":"A username to run perf top-down analysis. Some application requires root permission to get running.",
   "user":"root",
   "_root_dir":"Root directory mounted into the docker container home, for instance, $tmp_dir as a root home directory",
-  "root_dir":"/home/$USER",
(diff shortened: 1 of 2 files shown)


workloads/spec2026/spec2026_perf.json
```diff
@@ -4,7 +4,9 @@
   "_user":"A username to run perf top-down analysis. Some application requires root permission to get running.",
   "user":"root",
   "_root_dir":"Root directory mounted into the docker container home, for instance, $tmp_dir as a root home directory",
-  "root_dir":"/home/$USER",
+  "root_dir":"/home/$USER/root_dir",
```
5surim, line 7: Why should it explicitly be root_dir?

estodi: No strong reason, just to indicate it does not have to be home dir, but we can revert this. Should we?

5surim (approved):
LGTM

PR #439, opened 2026-08-26
Fix `--trace` removes unrelated workload_db fields

estodi:
Previously, `--trace` mode stored results in workload_db while ignoring any unrelated existing fields, such as `"performance"`. This PR addresses the issue by appending the existing non-trace fields after the trace-related fields.


5surim (approved):
LGTM

estodi:
Thank you!

PR #440, opened 2026-09-01
Add workload to slurm log file names in trace mode

estodi:
In parallel segments trace mode, all log files went to a single log directory in `job_%j.out` formats, which made it difficult to find the exact log file for a specific workload. This PR adds small fixes to address this issue.


hlitz (approved):
lgtm

PR #442, opened 2026-09-01
Weekly sweep: check out a fresh scarab-infra too, and run Sunday night

hlitz:
Follow-up to #437, rebased on current main.

## Why the 2026-08-30 sweep mailed FAILED

All three modes died in `--build-scarab`:

```
pin_exec.cpp:181:10: error: 'INS_HasScatteredMemoryAccess' was not declared in this scope
```

The sweep resets its *Scarab* clone to `origin/main` but ran whatever
*scarab-infra* it was started from — cron starts it in a shared working tree,
which sat on a branch predating #435 (the SDE 9.44.0 / PIN 3.31 image). A
PIN-3.15 image was asked to build a Scarab main that needs PIN 3.31. Reproduced
exactly: that checkout resolves to `allbench_traces:6e61bf2f9b9b`
(`PIN_ROOT=/tmp_home/pin-3.15-…`); main resolves to `b8303c9d1e27` (SDE 9.44.0).

## Check out both repos, like CI

`refresh_infra()` does for infra what `refresh_scarab()` already does for
Scarab: clone/reset `/soe/hlitz/git/scarab-infra-weekly` to `origin/main` and
re-exec the sweep from it before anything else runs. Whatever checkout cron
starts only supplies that first line of code; the descriptors, `sci`, run
scripts and image tag all come from the fresh main — the same shape as the CI
job, which checks out both repos per run.

Guards, because re-execing into code you have not read is how you get a loop or
a silent death:

- `--no-self-update` runs a checkout as-is, and its mail carries a "not main"
  warning naming the sha and how far behind it is.
- The re-exec is skipped when the script on main does not yet understand that
  flag — otherwise an older main dies on an unknown argument (argparse exits
  before the crash-mail handler exists) or re-execs forever.

`refresh_scarab()` also cleans `src/pin`: the PIN tool's `.d` files name the PIN
path of the image that built them, so after an image switch make looks for
`/tmp_home/pin-3.15-…/algorithm` and stops. Scoped to `src/pin` so DynamoRIO is
not rebuilt every week.

## Schedule

The host is UTC and this cron build has no `CRON_TZ` (absent from the binary
and the man page), so `0 0 * * 0` fired **Saturday 17:00 Pacific** — that is
the "Aug 29, 5:02 PM" timestamp on the failure mail, not Sunday night. Now
`0 5 * * 1`: Monday 05:00 UTC = Sunday 22:00 PDT / 21:00 PST. The cron line
also drops its `cd`-and-git preamble, since the script does the checkout now.

## Verified

- Re-exec against a stand-in origin carrying this code: clone created, reset,
  `running from … at fed4e84`, sweep continues there — no loop, no argparse
  error.
- Against a main without the flag: declines to re-exec, logs and mails
  `predates --no-self-update`.
- Mail paths still fire after the rebase: FAILED, (partial), success, CRASHED,
  SKIPPED, and nothing under `--dry-run`.
- After cleaning `src/pin`, `pin_exec` compiles and links against SDE 9.44.0 in
  the current image.

## Not fixed here

`sci` names the build container `allbench_traces_<user>_scarab_build`
(`utilities.py:404`) with no per-run suffix, and every build starts with
`docker rm -f` on that name. Two concurrent `--build-scarab` runs by the same
user therefore destroy each other's container, and the loser dies mid-build
with no error text at all — just a truncated log. `docker events` shows the
`kill`/`die`/`destroy` landing on an in-flight build. The sim containers
already carry a unique suffix; the build container should too. Worth its own PR.

### DIFF
--- a/scripts/weekly_perf_sweep.cron
+++ b/scripts/weekly_perf_sweep.cron
@@ -4,8 +4,18 @@
 #
 # Needs the scarabinfra conda env (pandas/matplotlib) and gh on PATH; cron's
 # PATH is minimal, so both are set explicitly below.
+#
+# The host runs UTC and this cron build has no CRON_TZ, so the schedule is
+# written in UTC: Monday 05:00 UTC is Sunday 22:00 Pacific (21:00 under PST),
+# i.e. Sunday night here. The old "Sunday 00:00 UTC" fired Saturday afternoon.
+#
+# It runs from a clone of its own, reset to main first. The sweep resets its
+# Scarab clone to origin/main, and pairing that with whatever branch someone
+# left checked out in a shared working tree is how 2026-08-30 came to build a
+# PIN-3.31 Scarab inside a PIN-3.15 image. Create the clone once with:
+#   git clone git@github.com:litz-lab/scarab-infra.git /soe/hlitz/git/scarab-infra-weekly
 SHELL=/bin/bash
 PATH=/soe/hlitz/miniconda3/envs/scarabinfra/bin:/usr/local/bin:/usr/bin:/bin
 MAILTO=""
 
-0 0 * * 0 cd /soe/hlitz/git/scarab-infra && python3 scripts/weekly_perf_sweep.py >> /soe/hlitz/logs/weekly_perf_sweep.log 2>&1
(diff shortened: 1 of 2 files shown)


hlitz:
Superseded by #441, which now carries the same two commits rebased on current main (77f5445). #441 was the stale branch: it still held the pre-squash originals of the work #437 already merged, which is what conflicted.

PR #445, opened 2026-09-03
Add application mount in `--perf`

estodi:
This PR adds application mount support to perf mode, following the recent changes in exec and trace modes.
Each workload runs in its own local directory and does not pollute the read-only mounted application. `stage_local_rundir` function is moved out of run_simpoint_trace.py into a standalone common script, so that it can also be used by perf mode.


hlitz (approved):
LGTM

